PR #25, opened 2019-04-01
Split gas tests

### LINKED ISSUE #10: Decouple HexSumTree gas analysis from tests
Currently running `test/hex-tree.js` takes a lot of time given that we are adding a lot of entries to the tree to test the gas usage.

We should be doing this in a separate script and not in the unit tests.

bingen:
And add more gas tests for SumTree Sortition lib.

Fixes #10.

### DIFF
--- a/contracts/lib/HexSumTree.sol
+++ b/contracts/lib/HexSumTree.sol
@@ -86,7 +86,7 @@ library HexSumTree {
         uint256 checkingNode;
         uint256 nodeSum;
         for (; checkingLevel > INSERTION_DEPTH; checkingLevel--) {
-            for (; child < CHILDREN; child++) {
+            for (child = 0; child < CHILDREN; child++) {
                 // shift the iterator and add it to node 0x00..0i00 (for depth = 3)
                 uint256 iterator = child << shift;
                 checkingNode = parentNode + iterator;
--- a/contracts/test/HexSumTreePublic.sol
+++ b/contracts/test/HexSumTreePublic.sol
@@ -19,6 +19,10 @@ contract HexSumTreePublic {
         emit LogKey(bytes32(tree.insert(v)));
     }
 
+    function insertNoLog(uint256 v) public {
+        tree.insert(v);
+    }
+
     function insertMultiple(uint256 v, uint256 number) public {
         for (uint256 i = 0; i < number; i++) {
             tree.insert(v);
@@ -34,15 +38,30 @@ contract HexSumTreePublic {
         emit LogKey(bytes32(key));
     }
 
+    function removeNoLog(uint256 key) public {
+        tree.set(key, 0);
+    }
+
     function removeMultiple(uint256 key, uint256 number) public {
         for (uint256 i = 0; i < number; i++) {
             tree.set(key + i, 0);
         }
     }
 
+    function setNextKey(uint256 key) public {
+        tree.nextKey = key;
+    }
+
     function sortition(uint256 v) public view returns (uint256) {
         var (k,) = tree.sortition(v);
         return uint256(k);
+    }
+
+    function multiRandomSortition(uint256 number) public {
+        for (uint256 i = 0; i < number; i++) {
+            bytes32 seed = keccak256(abi.encodePacked(block.number, i));
+            tree.randomSortition(uint256(seed));
+        }
     }
 
     function get(uint256 l, uint256 key) public view returns (uint256) {
--- a/contracts/test/HexSumTreePublic.sol
+++ b/contracts/test/HexSumTreePublic.sol
@@ -10,69 +10,76 @@ contract HexSumTreePublic {
 
     event LogKey(bytes32 k);
     event LogRemove(bytes32 k);
+    event GasConsumed(uint256 gas);
+
+    modifier profileGas {
+        uint256 initialGas = gasleft();
+        _;
+        emit GasConsumed(initialGas - gasleft());
+    }
 
     function init() public {
         tree.init();
     }
 
-    function insert(uint256 v) public {
+    function insert(uint256 v) external profileGas {
         emit LogKey(bytes32(tree.insert(v)));
     }
 
-    function insertNoLog(uint256 v) public {
+    function insertNoLog(uint256 v) external profileGas {
         tree.insert(v);
     }
 
-    function insertMultiple(uint256 v, uint256 number) public {
+    function insertMultiple(uint256 v, uint256 number) external profileGas {
         for (uint256 i = 0; i < number; i++) {
             tree.insert(v);
         }
     }
 
-    function set(uint256 key, uint256 value) public {
+    function set(uint256 key, uint256 value) external profileGas {
         tree.set(key, value);
     }
 
-    function remove(uint256 key) public {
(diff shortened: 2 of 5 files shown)


contracts/test/HexSumTreePublic.sol
```diff
@@ -19,6 +19,10 @@ contract HexSumTreePublic {
         emit LogKey(bytes32(tree.insert(v)));
     }
 
+    function insertNoLog(uint256 v) public {
```
izqui: Making these function external would be a bit more efficient which is better for more accurate gas benchmarks

test/hex-tree-gas.js
```diff
@@ -0,0 +1,253 @@
...
+    await tree.init()
+  })
+
+  const assertBN = (bn, n, m) => {
+    assert.equal(bn.toNumber(), n, m)
+  }
+
+  const logTreeState = async () => {
+    //console.log((await tree.getState()).map(x => x.toNumber()))
+    const state = await tree.getState()
```
izqui:
```suggestion
    const [ depth, nextKey ] = await tree.getState()
```

test/hex-tree-gas.js
```diff
@@ -0,0 +1,253 @@
...
+    console.log(`Tree depth:    ${state[0]}`);
+    console.log(`Tree next key: ${state[1].toNumber().toLocaleString()}`);
```
izqui: What's the reason for using `toLocaleString()`?

test/hex-tree-gas.js
```diff
@@ -0,0 +1,253 @@
+const assertRevert = require('./helpers/assert-revert')
+
+const HexSumTreePublic = artifacts.require('HexSumTreePublic')
+
+const CHILDREN = 16
+
+contract('Hex Sum Tree', (accounts) => {
```
izqui:
```suggestion
contract('Hex Sum Tree (Gas analysis)', (accounts) => {
```

test/hex-tree-gas.js
```diff
@@ -0,0 +1,253 @@
...
+    console.log(`Tree depth:    ${state[0]}`);
+    console.log(`Tree next key: ${state[1].toNumber().toLocaleString()}`);
+    console.log(`Tree total sum: `, (await tree.totalSum()).toNumber().toLocaleString())
+  }
+
+  const logSortitionGas = async (value) => {
+    console.log(`Sortition ${value} gas:`, (await tree.sortition.estimateGas(value)).toLocaleString())
+  }
+
+  const logSortitionHex = async (value) => {
```
izqui: Function is unused

contracts/lib/HexSumTree.sol
```diff
@@ -86,7 +86,7 @@ library HexSumTree {
         uint256 checkingNode;
         uint256 nodeSum;
         for (; checkingLevel > INSERTION_DEPTH; checkingLevel--) {
-            for (; child < CHILDREN; child++) {
+            for (child = 0; child < CHILDREN; child++) {
```
izqui: How weren't we having issues with this before?

test/hex-tree-gas.js
```diff
@@ -0,0 +1,253 @@
...
+    console.log('|---------|' + '-'.padStart(COL_SIZE + 2, '-') + '|' + '-'.padStart(COL_SIZE + 2, '-') + '|')
+    console.log('| Min     |', formatDivision(min / batchSize, COL_SIZE), '|', formatDivision((min - TX_GAS) / batchSize, COL_SIZE), '|')
+    console.log('| Max     |', formatDivision(max / batchSize, COL_SIZE), '|', formatDivision((max - TX_GAS) / batchSize, COL_SIZE), '|')
+    console.log('| Average |', formatDivision(avg / batchSize, COL_SIZE), '|', formatDivision((avg - TX_GAS) / batchSize, COL_SIZE), '|')
+    console.log()
+  }
+
+  it('inserts one node', async () => {
+    const r = await tree.insertNoLog(10)
+    const insertGas = r.receipt.cumulativeGasUsed
```
izqui: We should use `gasUsed`, `cumulativeGasUsed` includes the gas used by other transactions in the block that came before this one. This is not an issue if there is always just one transaction per block, but we should fix it.

test/hex-tree-gas.js
```diff
@@ -0,0 +1,253 @@
...
+  const logTreeState = async () => {
+    //console.log((await tree.getState()).map(x => x.toNumber()))
+    const state = await tree.getState()
+    console.log(`Tree depth:    ${state[0]}`);
+    console.log(`Tree next key: ${state[1].toNumber().toLocaleString()}`);
+    console.log(`Tree total sum: `, (await tree.totalSum()).toNumber().toLocaleString())
+  }
+
+  const logSortitionGas = async (value) => {
+    console.log(`Sortition ${value} gas:`, (await tree.sortition.estimateGas(value)).toLocaleString())
```
izqui: I have done a few tests and using `estimateGas` seems to return the exact same value as the gas used by performing the call as a transaction.

test/hex-tree-gas.js
izqui: This test file is quite slow and I don't see any assertions that aren't done somewhere else. We should consider skipping these tests in most cases, unless you are explicitly profiling gas.

test/hex-tree-gas.js
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  const formatDivision = (result, colSize) => {
+    return Math.round(result).toLocaleString().padStart(colSize, ' ')
+  }
+  const logGasStats = (title, gasArray, batchSize = 1) => {
+    const COL_SIZE = 7
+    console.log(title)
+    console.log('Size:   ', gasArray.length)
+    const TX_GAS = 21000
```
izqui: Not a really big difference, but if we wanted to truly know how much gas a function uses we could use something like this:

```sol
modifier profileGas {
    uint256 initialGas = gasleft();
    _;
    emit GasConsumed(initialGas - gasleft());
  }
```

This is the most accurate way I can think of to properly profile how much gas a function really uses.

test/hex-tree-gas.js
bingen: Yes, I did some tests too and had the same result. As rule of thumb, I'm using `estimateGas` when I don't need the transaction to be executed (e.g., when it's a read call), as the code is shorter, and the real call and the receipt when I need the state change.

test/hex-tree-gas.js
bingen: Yes, but it's so useful to have it handy when things go wrong! ;-) Besides, as we talked offline, these are not going to be the final tests, so meanwhile I would like to keep it.

test/hex-tree-gas.js
bingen: To have thousands separators. Is there a better way of doing it?

test/hex-tree-gas.js
izqui: That makes sense. Gas estimations tend to break when there are refunds

contracts/lib/HexSumTree.sol
bingen: Added test for it: a6559ab

Issue #16, opened 2019-03-23
Court: Implement more getters to fetch state


izqui:
Right now most of the Court's state can be reconstructed checking the logs of the contract, but we should provide some getters (`external view`s) that allow for quickly fetching the state of a dispute, adjudication round and juror accounts.

facuspagnuolo:
Multiple getters were already added to fetch configs, dispute, round, and appeal states. However, many state variables are still `internal` without a getter, we should probably want to either provide a getter if we don't reach the contract size limit or make them `pubilc`

PR #30, opened 2019-04-17
Issue #16. Adding basic getters for dispute, round and account, including tests.

willjgriff:
Fixes issue #16 

This is my first PR with you guys, hoping to work on some of the other more basic issues as well. 

I added tests although I understand if they're overkill for getters in which case I can remove them.

I assumed it wouldn't be useful to expose the AdjudicationRound's `votes` and `rulingVotes` or the Account's `update` struct in separate getters as they seem like things that only need to be used internally.

I considered adding a `getUnlockedAccountBalance(address _accountAddress, address _tokenAddress)` which would return `accounts[_accountAddress].balances[_tokenAddress]` which could be used to find the values needed for the function `withdraw(ERC20 _token, uint256 _amount)`. However, I can't see any way of staking or otherwise sending tokens besides the juror token to this contract, and see there's an `unlockedBalanceOf(address _addr)` function which does the above for the juror token, so figured it's probably not useful.

Will gladly add any of the above if it's expected they could be useful though.

### DIFF
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -844,8 +844,7 @@ contract Court is ERC900, ApproveAndCallFallBack {
         view
         returns (uint8 winningRuling, uint64 draftTerm, uint64 jurorNumber, address triggeredBy, bool settledPenalties, uint256 slashedTokens)
     {
-        Dispute storage dispute = disputes[_disputeId];
-        AdjudicationRound storage round = dispute.rounds[_roundId];
+        AdjudicationRound storage round = disputes[_disputeId].rounds[_roundId];
         return (round.winningRuling, round.draftTerm, round.jurorNumber, round.triggeredBy, round.settledPenalties, round.slashedTokens);
     }
 
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -834,6 +834,30 @@ contract Court is ERC900, ApproveAndCallFallBack {
         return account.balances[jurorToken].sub(account.atStakeTokens);
     }
 
+    function getDispute(uint256 _disputeId) external view returns (address subject, uint8 possibleRulings, DisputeState state) {
+        Dispute storage dispute = disputes[_disputeId];
+        return (dispute.subject, dispute.possibleRulings, dispute.state);
+    }
+
+    function getAdjudicationRound(uint256 _disputeId, uint256 _roundId)
+        external
+        view
+        returns (uint8 winningRuling, uint64 draftTerm, uint64 jurorNumber, address triggeredBy, bool settledPenalties, uint256 slashedTokens)
+    {
+        Dispute storage dispute = disputes[_disputeId];
+        AdjudicationRound storage round = dispute.rounds[_roundId];
+        return (round.winningRuling, round.draftTerm, round.jurorNumber, round.triggeredBy, round.settledPenalties, round.slashedTokens);
+    }
+
+    function getAccount(address _accountAddress)
+        external
+        view
+        returns (AccountState state, uint64 fromTerm, uint64 toTerm, uint256 atStakeTokens, uint256 sumTreeId)
+    {
+        Account storage account = accounts[_accountAddress];
+        return (account.state, account.fromTerm, account.toTerm, account.atStakeTokens, account.sumTreeId);
+    }
+
     function _editUpdate(AccountUpdate storage update, uint256 _delta, bool _positive) internal {
         (update.delta, update.positive) = _signedSum(update.delta, update.positive, _delta, _positive);
     }
--- a/test/court-disputes.js
+++ b/test/court-disputes.js
@@ -110,10 +110,10 @@ contract('Court: Disputes', ([ poor, rich, governor, juror1, juror2, juror3, arb
     await assertEqualBN(this.court.totalStakedFor(juror3), juror3Stake, 'juror3 stake')
   })
 
-  // it('can encrypt votes', async () => {
-  //   const ruling = 10
-  //   assert.equal(await this.court.encryptVote(ruling, SALT), encryptVote(ruling))
-  // })
+  it('can encrypt votes', async () => {
+    const ruling = 10
+    assert.equal(await this.court.encryptVote(ruling, SALT), encryptVote(ruling))
+  })
 
   context('activating jurors', () => {
     const passTerms = async terms => {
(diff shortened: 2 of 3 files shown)


contracts/Court.sol
```diff
@@ -834,6 +834,30 @@ contract Court is ERC900, ApproveAndCallFallBack {
...
+        return (dispute.subject, dispute.possibleRulings, dispute.state);
+    }
+
+    function getAdjudicationRound(uint256 _disputeId, uint256 _roundId)
+        external
+        view
+        returns (uint8 winningRuling, uint64 draftTerm, uint64 jurorNumber, address triggeredBy, bool settledPenalties, uint256 slashedTokens)
+    {
+        Dispute storage dispute = disputes[_disputeId];
+        AdjudicationRound storage round = dispute.rounds[_roundId];
```
bingen: I guess we can do this in just one line, as we don't actually need `dispute`.

contracts/Court.sol
```diff
@@ -834,6 +834,30 @@ contract Court is ERC900, ApproveAndCallFallBack {
...
+        return (round.winningRuling, round.draftTerm, round.jurorNumber, round.triggeredBy, round.settledPenalties, round.slashedTokens);
```
bingen: Isn't this line too long? Have you tried "linting" it?

test/court-disputes.js
```diff
@@ -141,9 +141,33 @@ contract('Court: Disputes', ([ poor, rich, governor, juror1, juror2, juror3, arb
...
+
+        assert.strictEqual(dispute[0], arbitrable, 'COURT_INCORRECT_DISPUTE_SUBJECT')
+        assertEqualBN(dispute[1], rulings, 'COURT_INCORRECT_DISPUTE_RULINGS')
+        assertEqualBN(dispute[2], 0, 'COURT_INCORRECT_DISPUTE_STATE')
+      })
+
+      it('gets the correct dispute round details', async () => {
+        const expectedWinningRuling = 0
+        const exepctedRoundPenalties = false
+        const expectedRoundSlashing = 0
```
bingen: I wonder it would be useful to get some "non-default" values here too.

test/court-disputes.js
```diff
@@ -141,9 +141,33 @@ contract('Court: Disputes', ([ poor, rich, governor, juror1, juror2, juror3, arb
...
 
       beforeEach(async () => {
-        await assertLogs(this.court.createDispute(arbitrable, rulings, jurors, term), NEW_DISPUTE_EVENT)
+        await assertLogs(this.court.createDispute(arbitrable, rulings, jurors, term, { from: poor }), NEW_DISPUTE_EVENT)
       })
 
+      it('gets the correct dispute details', async () => {
+        const dispute = await this.court.getDispute(disputeId)
+
+        assert.strictEqual(dispute[0], arbitrable, 'COURT_INCORRECT_DISPUTE_SUBJECT')
```
izqui: The assertion messages that you have chosen here are have the same format as the smart contract errors (which are meant to be identifiers), but for the assertions we tend to just use more descriptive messages.

```suggestion
        assert.strictEqual(dispute[0], arbitrable, 'incorrect dispute subject')
```

test/court-disputes.js
```diff
@@ -141,9 +141,33 @@ contract('Court: Disputes', ([ poor, rich, governor, juror1, juror2, juror3, arb
       const firstRoundId = 0
 
       beforeEach(async () => {
-        await assertLogs(this.court.createDispute(arbitrable, rulings, jurors, term), NEW_DISPUTE_EVENT)
+        await assertLogs(this.court.createDispute(arbitrable, rulings, jurors, term, { from: poor }), NEW_DISPUTE_EVENT)
       })
 
+      it('gets the correct dispute details', async () => {
+        const dispute = await this.court.getDispute(disputeId)
```
izqui: Arrays from the getters can be destructured to avoid having magic numbers for accessing array items:

```
const [
  disputeSubject,
  disputeRulings,
  disputeState
] = await this.court.getDispute(disputeId)
```

test/court-disputes.js
```diff
@@ -141,9 +141,33 @@ contract('Court: Disputes', ([ poor, rich, governor, juror1, juror2, juror3, arb
...
+
+        assert.strictEqual(dispute[0], arbitrable, 'COURT_INCORRECT_DISPUTE_SUBJECT')
+        assertEqualBN(dispute[1], rulings, 'COURT_INCORRECT_DISPUTE_RULINGS')
+        assertEqualBN(dispute[2], 0, 'COURT_INCORRECT_DISPUTE_STATE')
```
izqui: Nitpick. A map from the dispute state name to the actual state integer representation can be implemented so states can be referred by name:

```
const DISPUTE_STATES = {
  PRE_DRAFT: 0,
  ADJUDICATING: 1,
  EXECUTED: 2,
  DISMISSED: 3
}
```
This would allow to modify the assertion to:
```suggestion
        assertEqualBN(dispute[2], DISPUTE_STATES.PRE_DRAFT, 'COURT_INCORRECT_DISPUTE_STATE')
```

test/court-disputes.js
```diff
@@ -141,9 +141,33 @@ contract('Court: Disputes', ([ poor, rich, governor, juror1, juror2, juror3, arb
...
+
+      it('gets the correct dispute round details', async () => {
+        const expectedWinningRuling = 0
+        const exepctedRoundPenalties = false
+        const expectedRoundSlashing = 0
+
+        const round = await this.court.getAdjudicationRound(disputeId, firstRoundId)
+
+        assertEqualBN(round[0], expectedWinningRuling, 'COURT_INCORRECT_ROUND_RULING')
+        assertEqualBN(await round[1], term, 'COURT_INCORRECT_ROUND_TERM')
```
izqui: I don't think these 2 `await` are needed

contracts/Court.sol
willjgriff: Yup, sounds good.

contracts/Court.sol
willjgriff: I've run the linter on it and it doesn't warn for length, it's 129 chars and the linter's set to 145. I can still change it. Is there any way you would recommend wrapping this/splitting it up if it is too long?

test/court-disputes.js
willjgriff: That makes sense, I noticed a mix of assert message formats but I missed that association. Thanks!

test/court-disputes.js
willjgriff: I've still got bits to learn about JS, this is much nicer, thanks.

test/court-disputes.js
willjgriff: Great idea, will do.

test/court-disputes.js
willjgriff: Good spot, removed.

test/court-disputes.js
willjgriff: I've set it up to check for non-default values now. However, it is now testing many things, beyond just the getter, since I haven't mocked anything. If you know of a better way, or if you think I should mock some stuff instead then I can change it.

contracts/Court.sol
bingen: It's fine then.

Issue #17, opened 2019-03-23
Enforce a max size for updateQueue and egressQueue for any term


izqui:
If any of the queues for a term is too large, this could result in the Court bricking due to `heartbeat()` always going OOG as the gas it needs to process its queues is greater than the block gas limit.

bingen:
We don't have queues anymore, so closing this one.

PR #31, opened 2019-04-18
Issue #17. Set max queue size.

willjgriff:
Fixes issue #17 

I did some benchmarking to find out the upper limit of each queue before OOG occurs on a single heartbeat (I haven’t done benchmarking executing multiple heartbeats in the same transaction).

**Gas limit: 10e6**
Update queue limit (egress queue empty): 195 addresses
Egress queue limit (update queue empty): 149 addresses
Egress queue filled including a mock account update increasing processing (update queue empty): 145 addresses

**Gas limit 6e6**
Update queue limit (egress queue empty): 115 addresses
Egress queue limit (update queue empty): 89 addresses
Egress queue filled including a mock account update increasing processing (update queue empty): 87 addresses

This suggests the total upper limit is 87 addresses. Divide this by 2 as each list is always added to/removed from at the same rate gives 43. This could be increased slightly as the half that are in the update queue use less gas. However, again note this is executing a single heartbeat, not multiple heartbeats in the same transaction. Benchmarking code can be seen [here](https://github.com/willjgriff/aragon-court/compare/master...willjgriff:issue_17_set_max_queue_sizes_benchmarking)

I’m not sure about the best way to test it, the current test can take up to 30 seconds, perhaps that's acceptable. I've also removed from the test the potential extra updates accounts could have due to slashing. I can remove the test/change it. Perhaps I should mock the QUEUE_MAX_SIZE in the MockCourt contract to a smaller value to make it run quicker? I haven’t tested the individual functions require statements.

Edit: I updated the benchmarking tests code to use a gas limit of 6e6 instead of 10e6 and included the results above.

### DIFF
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -177,7 +177,7 @@ contract Court is ERC900, ApproveAndCallFallBack {
     uint8 public constant MIN_RULING_OPTIONS = 2;
     uint8 public constant MAX_RULING_OPTIONS = MIN_RULING_OPTIONS;
     address internal constant BURN_ACCOUNT = 0xdead;
-    uint8 private constant QUEUE_MAX_SIZE = 72;
+    uint8 private constant QUEUE_MAX_SIZE = 43;
 
     event NewTerm(uint64 term, address indexed heartbeatSender);
     event NewCourtConfig(uint64 fromTerm, uint64 courtConfigId);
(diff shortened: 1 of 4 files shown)


willjgriff:
The test is failing due to the test requiring a ganache-cli instance with 150 accounts. Eg `ganache-cli -a 150`. I will investigate this properly soon.

izqui:
Thanks for this contribution @willjgriff. We are currently in the process of deciding whether to remove these queues for the most part if checkpointing gets added (see #28).

PR #39, opened 2019-05-14
Split sumtree

### LINKED ISSUE #8: Explore ways to split Court into multiple contracts
For simplicity, the Court implementation has been done in just one contract as the state of the contract is quite interdependent between most parts of the mechanism. An idea for separating concerns would be to split the general court state from the dispute management and adjudication.

We are still well below the block gas limit (at the moment ~5.5m gas), but as we keep implementing new functionality, we will start approaching the limit.

bingen:
Closes #8 

TODO:
- [x] Protect HexSumTreeWrapper so only owner (Court) can alter the tree


contracts/HexSumTreeWrapper.sol
```diff
@@ -0,0 +1,106 @@
+pragma solidity ^0.4.24;
+
+import "./lib/HexSumTree.sol";
+
+
+contract HexSumTreeWrapper {
```
bingen: TODO: protect this so only owner (Court) can alter the tree.

facuspagnuolo:
Closing this PR since `development` branch was already updated with these commits manually

PR #45, opened 2019-05-27
Final appeal round

### LINKED ISSUE #18: Implement a limit for how many regular appeals can happen for a dispute

### LINKED ISSUE #19: Implement final appeal
After the maximum number of appeals has been reached, if a new appeal is created, the final ruling by the Court should be decided with a stake-weighted vote of all active jurors.

bingen:
Closes #18 
Closes #19 

TODO:

- [X] Fees.
- [ ] Lock juror stakes.

### DIFF
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -19,6 +19,8 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
     using SafeMath for uint256;
 
     uint256 internal constant MAX_JURORS_PER_BATCH = 10; // to cap gas used on draft
+    uint256 internal constant MAX_DRAFT_ROUNDS = 4; // before the final appeal
+    // TODO: move all other constants up here
 
     struct Account {
         mapping (address => uint256) balances; // token addr -> balance
@@ -515,8 +517,15 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         uint32 appealJurorNumber = 2 * currentRound.jurorNumber + 1; // J' = 2J + 1
         uint64 appealDraftTermId = termId + 1; // Appeals are drafted in the next term
 
-        // _newAdjudicationRound charges fees for starting the round
-        (uint256 roundId, uint32 voteId) = _newAdjudicationRound(_disputeId, appealJurorNumber, appealDraftTermId);
+        uint256 roundId;
+        uint32 voteId;
+        if (_roundId == MAX_DRAFT_ROUNDS - 1) { // roundId starts at 0
+            (roundId, voteId) = _finalAdjudicationRound(_disputeId, appealDraftTermId);
+        } else {
+            // _newAdjudicationRound charges fees for starting the round
+            (roundId, voteId) = _newAdjudicationRound(_disputeId, appealJurorNumber, appealDraftTermId);
+        }
+
         emit RulingAppealed(_disputeId, roundId, appealDraftTermId, voteId, appealJurorNumber);
     }
 
@@ -735,6 +744,11 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         uint256 roundId = vote.roundId;
         _checkAdjudicationState(disputeId, roundId, _state);
 
+        // for the final round
+        if (roundId == MAX_DRAFT_ROUNDS) {
+            return sumTree.getItemPast(accounts[_voter].sumTreeId, disputes[disputeId].rounds[roundId].draftTermId);
+        }
+
         return getJurorWeight(disputeId, roundId, _voter);
     }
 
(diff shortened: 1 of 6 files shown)


contracts/Court.sol
```diff
@@ -586,20 +640,23 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
+        uint256 _disputeId,
+        AdjudicationRound storage _round,
+        uint16 _penaltyPct
+    )
+        internal
+        returns (uint256)
+    {
+        for (uint256 i = finalAdjudicationRounds.length; i > 0; i--) {
+            if (finalAdjudicationRounds[i - 1].disputeId == _disputeId) {
+                return finalAdjudicationRounds[i - 1].lockedStakePerJuror * (_round.jurorNumber - _round.coherentJurors);
```
izqui: How would this take into account jurors that become over-leveraged?

contracts/Court.sol
```diff
@@ -616,25 +673,72 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
 
         jurorState.rewarded = true;
 
-        uint256 voteId = round.voteId;
-        uint8 winningRuling = _getWinningRuling(dispute);
-        uint256 coherentJurors = voting.getRulingVotes(voteId, winningRuling);
-        uint8 jurorRuling = voting.getCastVote(voteId, _juror);
+        uint256 coherentJurors = round.coherentJurors;
```
izqui: Do we need to prevent final round reward settlements from happening in this function as well?

contracts/Court.sol
```diff
@@ -616,25 +673,72 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
+
+        require(round.settledPenalties, ERROR_ROUND_NOT_SETTLED);
+
+        uint256 coherentJurors = round.coherentJurors;
+        uint8 jurorRuling = voting.getCastVote(round.voteId, _juror);
+        uint256 weight = _getJurorWeight(disputeId, MAX_DRAFT_ROUNDS, _juror);
+
+        if (jurorRuling == round.winningRuling) {
+            _assignTokens(jurorToken, _juror, round.slashedTokens * weight / coherentJurors);
+            _payFees(config.feeToken, _juror, config.jurorFee * round.jurorNumber * weight / coherentJurors, config.governanceFeeShare);
```
izqui: The rounding errors for distributing fees could be quite significant and a pretty big amount of fees would never be distributed.

contracts/Court.sol
```diff
@@ -19,6 +19,8 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
     using SafeMath for uint256;
 
     uint256 internal constant MAX_JURORS_PER_BATCH = 10; // to cap gas used on draft
+    uint256 internal constant MAX_DRAFT_ROUNDS = 4; // before the final appeal
```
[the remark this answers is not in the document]
bingen: Makes sense, but I see 2 problems:
- I'm afraid we'll hit stack too deep issue again in the constructor. Minor one, I'll figure it out.
- It could be a little bit messy for on-going disputes, the logic would be more complex. I have to review all the occurrences of it to make sure I don't break anything.
So I'm going to open an issue for this (#55), so we don't get stuck here, ok?

contracts/Court.sol
bingen: Hm, well, assuming typical 18 decimals are used, each time the max amount lost would be `1e-18` of a token, right? But still, yes, I'm aware, not only here, that's why I opened an issue for this: #47

izqui: The amount of fees collected depends on the `jurorNumber` (which for the final round is calculated with `totalStake / minStake`), but each juror's weight is `stake / minStake`. I think the sum of the rest of the `stake / minStake` division for all jurors will be 'ghost stake' that is used for calculating the fee amount, but isn't actually distributed.

For example, there are only 2 jurors with 15 tokens each and the `minStake` is 10 tokens. Each juror will have a weight of 1, even though the total stake is 30 and therefore the `jurorNumber` is 3. If both jurors are coherent, when distributing the fees each juror will get 1/3 of the total reward, leaving 1/3 stuck.

bingen: Fair enough, although I think that error then belongs to here: https://github.com/aragon/aragon-court/blob/issue_19_final_appeal/contracts/Court.sol#L862

contracts/Court.sol
```diff
@@ -19,28 +19,32 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
+    // TODO: move all other constants up here
 
     struct Account {
         mapping (address => uint256) balances; // token addr -> balance
         // when deactivating, balance becomes available on next term:
         uint64 deactivationTermId;
+        uint64 nextFinalAdjudicationRoundToSettle; // id in finalAdjudicationRounds array of the next pending final appeal round to settle
```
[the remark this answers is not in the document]
bingen: Thanks for catching this! I will push fixes in a while, but my initial approach, in order to avoid crazy complexity, is to forbid re-activations if there are outstanding final rounds, so jurors must settle them first.

contracts/Court.sol
bingen: Good catch!

bingen: Well, actually, I think weight here would be always 0, so it would revert, but I'm adding a sanity check.

contracts/Court.sol
bingen: Actually: https://github.com/aragon/aragon-court/blob/issue_19_final_appeal/contracts/Court.sol#L901
:wink:

bingen:
Closed in favor of #56

PR #56, opened 2019-06-12
Court: implement final appeal round

### LINKED ISSUE #18: Implement a limit for how many regular appeals can happen for a dispute

### LINKED ISSUE #19: Implement final appeal

bingen:
Closes #18 and #19

### DIFF
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -586,8 +586,10 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
             slashedTokens = _settleRegularRoundSlashing(round, voteId, config.penaltyPct, winningRuling);
             round.slashedTokens = slashedTokens;
             _payFees(config.feeToken, msg.sender, config.settleFee * round.jurorNumber, config.governanceFeeShare);
-        } else {
+        } else { // final round
+            // this was accounted for on juror's vote commit
             slashedTokens = round.slashedTokens;
+            // there's no settleFee in this round
         }
 
         round.settledPenalties = true;
@@ -595,7 +597,11 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         // No juror was coherent in the round
         if (coherentJurors == 0) {
             // refund fees and burn ANJ
-            _payFees(config.feeToken, round.triggeredBy, config.jurorFee * round.jurorNumber, config.governanceFeeShare);
+            uint256 jurorFee = config.jurorFee * round.jurorNumber;
+            if (_roundId == MAX_REGULAR_APPEAL_ROUNDS) {
+                jurorFee = _pct4(jurorFee / FINAL_ROUND_WEIGHT_PRECISION, config.finalRoundReduction);
+            }
+            _payFees(config.feeToken, round.triggeredBy, jurorFee, config.governanceFeeShare);
             _assignTokens(jurorToken, BURN_ACCOUNT, slashedTokens);
         }
 
@@ -611,7 +617,7 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         internal
         returns (uint256 slashedTokens)
     {
-        uint256 penalty = _pct4(jurorMinStake, _penaltyPct); // TODO: stack too deep
+        uint256 penalty = _pct4(jurorMinStake, _penaltyPct);
 
         uint256 votesLength = _round.jurors.length;
         uint64 slashingUpdateTermId = termId + 1;
@@ -619,7 +625,7 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         // should we batch this too?? OOG?
         for (uint256 i = 0; i < votesLength; i++) {
             address juror = _round.jurors[i];
-            uint256 weightedPenalty = penalty * _round.jurorSlotStates[juror].weight; // TODO: stack too deep
+            uint256 weightedPenalty = penalty * _round.jurorSlotStates[juror].weight;
             Account storage account = accounts[juror];
             account.atStakeTokens -= weightedPenalty;
 
@@ -666,7 +672,11 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         }
 
         CourtConfig storage config = courtConfigs[terms[round.draftTermId].courtConfigId]; // safe to use directly as it is a past term
-        _payFees(config.feeToken, _juror, config.jurorFee * jurorState.weight * round.jurorNumber / coherentJurors, config.governanceFeeShare);
+        uint256 jurorFee = config.jurorFee * jurorState.weight * round.jurorNumber / coherentJurors;
+        if (_roundId == MAX_REGULAR_APPEAL_ROUNDS) {
+            jurorFee = _pct4(jurorFee, config.finalRoundReduction);
+        }
+        _payFees(config.feeToken, _juror, jurorFee, config.governanceFeeShare);
 
         emit RewardSettled(_disputeId, _roundId, _juror);
     }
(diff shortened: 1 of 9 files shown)


contracts/Court.sol
```diff
@@ -612,20 +656,23 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
 
         if (slashedTokens > 0) {
-            _assignTokens(jurorToken, _juror, slashedTokens / coherentJurors);
+            _assignTokens(jurorToken, _juror, jurorState.weight * slashedTokens / coherentJurors);
         }
 
-        CourtConfig storage config = courtConfigs[terms[round.draftTermId].courtConfigId]; // safe to use directly as it is the current term
-        _payFees(config.feeToken, _juror, config.jurorFee * round.jurorNumber / coherentJurors, config.governanceFeeShare);
+        CourtConfig storage config = courtConfigs[terms[round.draftTermId].courtConfigId]; // safe to use directly as it is a past term
+        uint256 jurorFee = config.jurorFee * jurorState.weight * round.jurorNumber / coherentJurors;
```
izqui: Thinking that we could start storing the total juror fees collected for the round when it is created (for both regular and final rounds) so we don't have to calculate it in 3 different places (creating rounds, settling penalties and rewards).

contracts/Court.sol
```diff
@@ -711,23 +758,66 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
      * @return `_voter`'s weight
      */
-    function canCommit(uint256 _voteId, address _voter) external ensureTerm only(voting) returns (uint256) {
-        return _canPerformVotingAction(_voteId, _voter, AdjudicationState.Commit);
+    function canCommit(uint256 _voteId, address _voter) external ensureTerm only(voting) returns (uint256 weight) {
+        (uint256 disputeId, uint256 roundId) = _decodeVoteId(_voteId);
+        weight = _canPerformVotingAction(disputeId, roundId, _voter, AdjudicationState.Commit);
+
+        // if it's the final round, lock tokens
+        if (weight > 0) {
```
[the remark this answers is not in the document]
izqui: Also checking whether the round is the final one first will be a small optimization

contracts/Court.sol
```diff
@@ -756,9 +853,42 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
+        // apply final round discount
+        uint256 feeAmount = config.heartbeatFee + _pct4(jurorNumber * config.jurorFee / FINAL_ROUND_WEIGHT_PRECISION, config.finalRoundReduction);
+        if (feeAmount > 0) {
+            require(config.feeToken.safeTransferFrom(msg.sender, this, feeAmount), ERROR_DEPOSIT_FAILED);
+        }
+
+        roundId = _createRound(_disputeId, DisputeState.Adjudicating, _draftTermId, jurorNumber, jurorNumber);
+    }
+
+    function _createRound(
```
izqui: Related to the comment in `settleReward` about storing the total juror fee amount in the round struct, we could also perform the `transferFrom` of the fee amount within `_createRound`

contracts/Court.sol
```diff
@@ -756,9 +853,42 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
+    function _newFinalAdjudicationRound(
+        uint256 _disputeId,
+        uint64 _draftTermId
+    )
+        internal
+        returns (uint256 roundId, uint32 jurorNumber)
+    {
+        // TODO: check overflow? make it bigger? (uint32 feels too small)
+        // For instance, this would be enough to hold the total total token suply of 42.9M (more than current ANT one) in the tree if min stake was 10.
+        jurorNumber = uint32(FINAL_ROUND_WEIGHT_PRECISION * sumTree.totalSumPresent(termId) / jurorMinStake);
```
izqui: 🚨 The `jurorNumber` is being calculated based on the total amount of tokens in the tree at the start of the current term, but the round's draft term is the next term, and this is the term that is used to calculate the weights of the individual jurors. Therefore `round.jurorNumber` will not be the same as summing the weights of all the jurors that participate in the round.

I have been checking and I don't see how this could result in the Court distributing an excess of fees (given that calculations are always a form of `weight / coherentJurors`), but it is still a bit dangerous and we should check the implications better.

bingen: This is a really good point! And it seems to me that it goes beyond that, because even if used terms were the same, it could happen that not all jurors vote, so still the sum of weights wouldn't match `jurorNumber`.

contracts/Court.sol
```diff
@@ -547,29 +569,65 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
             _checkAdjudicationState(_disputeId, dispute.rounds.length - 1, AdjudicationState.Ended);
-        } else {
-            revert(ERROR_INVALID_DISPUTE_STATE);
         }
 
         uint256 voteId = _getVoteId(_disputeId, _roundId);
         uint8 winningRuling = voting.getWinningRuling(voteId);
-        CourtConfig storage config = courtConfigs[terms[round.draftTermId].courtConfigId]; // safe to use directly as it is the current term
-        // uint256 penalty = _pct4(jurorMinStake, config.penaltyPct); // TODO: stack too deep
+        round.winningRuling = winningRuling;
```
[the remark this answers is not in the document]
bingen: Good catch!

contracts/Court.sol
```diff
@@ -539,38 +564,81 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
         // Enforce that rounds are settled in order to avoid one round without incentive to settle
         // even if there is a settleFee, it may not be big enough and all jurors in the round are going to be slashed
         require(_roundId == 0 || dispute.rounds[_roundId - 1].settledPenalties, ERROR_PREV_ROUND_NOT_SETTLED);
         require(!round.settledPenalties, ERROR_ROUND_ALREADY_SETTLED);
 
+        uint256 voteId = _getVoteId(_disputeId, _roundId);
+        uint8 winningRuling;
         if (dispute.state != DisputeState.Executed) {
             _checkAdjudicationState(_disputeId, dispute.rounds.length - 1, AdjudicationState.Ended);
+            winningRuling = voting.getWinningRuling(voteId);
```
izqui: `voteId` should always be for the last round of the dispute. Unless the dispute has been executed (`executeRuling` always gets the winning ruling for the last round), the current implementation will still get the winning ruling for each round.

izqui: Also we could move this into a separate function that `executeRuling` uses as well, so if the winning ruling for the dispute has already been set we don't need to get it from voting again (and we can also skip checking the adjudication state of the last round).

```
function _ensureFinalRuling(uint256 _disputeId) internal returns (uint8 winningRuling) {

    if (dispute.winningRuling > 0) {
        return dispute.winningRuling; // winning ruling was already set
    }

    // ensure the last round adjudication period already ended
    uint256 lastRoundId = dispute.rounds.length - 1;
    _checkAdjudicationState(_disputeId, lastRound, AdjudicationState.Ended);

    uint256 voteId = _getVoteId(_disputeId, lastRoundId);
    uint8 winningRuling = voting.getWinningRuling(voteId);
    dispute.winningRuling = winningRuling;
    
    return winningRuling;
}
```

contracts/Court.sol
izqui: My worry is whether this could end up distributing more fees than collected, we need to make sure that doesn't happen for sure

contracts/Court.sol
bingen: What do you mean exactly? Here `jurorFee.weight` is different in each call. Do you mean caching the rest? (`config.jurorFee * round.jurorNumber / coherentJurors`) We would be saving `10` gas in math against the SSTORE, which would only be worth it on higher rounds. Besides, we could increase the probability of rounding problems due to the division. Or am I missing something?

contracts/Court.sol
bingen: > voteId should always be for the last round of the dispute

:man_facepalming: That fix was too fast, sorry. I hope it's right now.

contracts/Court.sol
```diff
@@ -570,13 +572,15 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         require(_roundId == 0 || dispute.rounds[_roundId - 1].settledPenalties, ERROR_PREV_ROUND_NOT_SETTLED);
         require(!round.settledPenalties, ERROR_ROUND_ALREADY_SETTLED);
 
+        uint8 winningRuling;
         if (dispute.state != DisputeState.Executed) {
```
izqui: Can't we just remove this `if` as if the dispute ruling has been executed `_ensureFinalRuling` will just return `dispute.winningRuling`

contracts/Court.sol
izqui: It would be `config.jurorFee * round.jurorNumber` (and in the case of the final round with the discount applied), this way we don't have to check the config and recalculate the fee amount every time.

bingen: Not sure about the gas, but the code will definitely look nicer.

bingen: Hm, now that I was doing it, we still need to get the `config` for the `feeToken` (unless we store it in `round` too, which seems  too much to me)

bingen:
Cloed in favor of #60

PR #57, opened 2019-07-02
Subscription fees

### LINKED ISSUE #51: Implement subscription fees

bingen:
Closes #51 

TODO:

- [X] Add events
- [x] Add function headers
- [x] Add tests
- [x] Add Court integration

### DIFF
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -158,6 +158,7 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
     uint8 internal constant MIN_RULING_OPTIONS = 2;
     uint8 internal constant MAX_RULING_OPTIONS = MIN_RULING_OPTIONS;
     address internal constant BURN_ACCOUNT = 0xdead;
+    uint256 internal constant MAX_UINT16 = uint16(-1);
     uint256 internal constant MAX_UINT32 = uint32(-1);
     uint64 internal constant MAX_UINT64 = uint64(-1);
 
@@ -197,10 +198,11 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
      * @param _feeToken The address of the token contract that is used to pay for fees.
      * @param _voting The address of the Commit Reveal Voting contract.
      * @param _sumTree The address of the contract storing de Sum Tree for sortitions.
-     * @param _jurorFee The amount of _feeToken that is paid per juror per dispute
-     * @param _heartbeatFee The amount of _feeToken per dispute to cover maintenance costs.
-     * @param _draftFee The amount of _feeToken per juror to cover the drafting cost.
-     * @param _settleFee The amount of _feeToken per juror to cover round settlement cost.
+     * @param _fees Array containing:
+     *        _jurorFee The amount of _feeToken that is paid per juror per dispute
+     *        _heartbeatFee The amount of _feeToken per dispute to cover maintenance costs.
+     *        _draftFee The amount of _feeToken per juror to cover the drafting cost.
+     *        _settleFee The amount of _feeToken per juror to cover round settlement cost.
      * @param _governor Address of the governor contract.
      * @param _firstTermStartTime Timestamp in seconds when the court will open (to give time for juror onboarding)
      * @param _jurorMinStake Minimum amount of juror tokens that can be activated
@@ -214,16 +216,14 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
         ICRVoting _voting,
         ISumTree _sumTree,
         ISubscriptions _subscriptions,
-        uint256 _jurorFee,
-        uint256 _heartbeatFee,
-        uint256 _draftFee,
-        uint256 _settleFee,
+        uint256[] _fees, // _jurorFee, _heartbeatFee, _draftFee, _settleFee
         address _governor,
         uint64 _firstTermStartTime,
         uint256 _jurorMinStake,
         uint64[3] _roundStateDurations,
         uint16 _penaltyPct,
-        uint16 _finalRoundReduction
+        uint16 _finalRoundReduction,
+        uint256[5] _subscriptionParams // _periodDuration, _feeAmount, _prePaymentPeriods, _latePaymentPenaltyPct, _governorSharePct
     ) public {
         termDuration = _termDuration;
         jurorToken = _jurorToken;
@@ -235,15 +235,13 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
 
         voting.setOwner(ICRVotingOwner(this));
         sumTree.init(address(this));
+        _initSubscriptions(_feeToken, _subscriptionParams);
 
         courtConfigs.length = 1; // leave index 0 empty
         _setCourtConfig(
             ZERO_TERM_ID,
             _feeToken,
-            _jurorFee,
-            _heartbeatFee,
-            _draftFee,
-            _settleFee,
+            _fees,
             _roundStateDurations,
             _penaltyPct,
             _finalRoundReduction
@@ -397,8 +395,9 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
         // TODO: Limit the max amount of terms into the future that a dispute can be drafted
         // TODO: Limit the max number of initial jurors
         // TODO: ERC165 check that _subject conforms to the Arbitrable interface
-        // TODO: Consider requiring that only the contract being arbitred can create a dispute
 
+        // TODO: require(address(_subject) == msg.sender, ERROR_INVALID_DISPUTE_CREATOR);
+        require(subscriptions.isUpToDate(address(_subject)), ERROR_SUBSCRIPTION_NOT_PAID);
         require(_possibleRulings >= MIN_RULING_OPTIONS && _possibleRulings <= MAX_RULING_OPTIONS, ERROR_INVALID_RULING_OPTIONS);
 
         uint256 disputeId = disputes.length;
@@ -1056,10 +1055,7 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
     function _setCourtConfig(
         uint64 _fromTermId,
         ERC20 _feeToken,
-        uint256 _jurorFee,
-        uint256 _heartbeatFee,
-        uint256 _draftFee,
-        uint256 _settleFee,
+        uint256[] _fees, // _jurorFee, _heartbeatFee, _draftFee, _settleFee
         uint64[3] _roundStateDurations,
         uint16 _penaltyPct,
         uint16 _finalRoundReduction
@@ -1081,10 +1077,10 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
 
         CourtConfig memory courtConfig = CourtConfig({
             feeToken: _feeToken,
-            jurorFee: _jurorFee,
-            heartbeatFee: _heartbeatFee,
-            draftFee: _draftFee,
-            settleFee: _settleFee,
+            jurorFee: _fees[0],
+            heartbeatFee: _fees[1],
+            draftFee: _fees[2],
+            settleFee: _fees[3],
             commitTerms: _roundStateDurations[0],
             revealTerms: _roundStateDurations[1],
             appealTerms: _roundStateDurations[2],
@@ -1097,6 +1093,22 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
         configChangeTermId = _fromTermId;
 
         emit NewCourtConfig(_fromTermId, courtConfigId);
+    }
+
+    function _initSubscriptions(ERC20 _feeToken, uint256[5] _subscriptionParams) internal {
+        require(_subscriptionParams[0] < MAX_UINT64); // _periodDuration
+        require(_subscriptionParams[3] < MAX_UINT16); // _latePaymentPenaltyPct
+        require(_subscriptionParams[4] < MAX_UINT16); // _governorSharePct
+        subscriptions.init(
+            ISubscriptionsOwner(this),
+            sumTree,
+            uint64(_subscriptionParams[0]), // _periodDuration
+            _feeToken,
+            _subscriptionParams[1],         // _feeAmount
+            _subscriptionParams[2],         // _prePaymentPeriods
+            uint16(_subscriptionParams[3]), // _latePaymentPenaltyPct
+            uint16(_subscriptionParams[4])  // _governorSharePct
+        );
     }
 
     function _time() internal view returns (uint64) {
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -832,7 +832,16 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
         return _canPerformVotingAction(disputeId, roundId, _voter, AdjudicationState.Reveal);
     }
 
-    function _canPerformVotingAction(uint256 _disputeId, uint256 _roundId, address _voter, AdjudicationState _state) internal view returns (uint256) {
+    function _canPerformVotingAction(
+        uint256 _disputeId,
+        uint256 _roundId,
+        address _voter,
+        AdjudicationState _state
+    )
+        internal
+        view
+        returns (uint256)
+    {
         _checkAdjudicationState(_disputeId, _roundId, _state);
 
         return _getJurorWeight(_disputeId, _roundId, _voter);
@@ -912,7 +921,8 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
         CourtConfig storage config = _courtConfigForTerm(_draftTermId);
         // number of jurors is the number of times the minimum stake is hold in the tree, multiplied by a precision factor for division roundings
         // besides, apply final round discount
-        uint256 feeAmount = config.heartbeatFee + _pct4(jurorNumber * config.jurorFee / FINAL_ROUND_WEIGHT_PRECISION, config.finalRoundReduction);
+        uint256 feeAmount = config.heartbeatFee +
+            _pct4(jurorNumber * config.jurorFee / FINAL_ROUND_WEIGHT_PRECISION, config.finalRoundReduction);
         if (feeAmount > 0) {
             require(config.feeToken.safeTransferFrom(msg.sender, this, feeAmount), ERROR_DEPOSIT_FAILED);
         }
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -241,7 +241,7 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
 
         voting.setOwner(ICRVotingOwner(this));
         sumTree.init(address(this));
-        _initSubscriptions(_feeToken, _subscriptionParams);
+        _initSubscriptions(_feeToken, _subscriptionParams, _sumTree);
 
         courtConfigs.length = 1; // leave index 0 empty
         _setCourtConfig(
@@ -1111,13 +1111,13 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
         emit NewCourtConfig(_fromTermId, courtConfigId);
     }
 
-    function _initSubscriptions(ERC20 _feeToken, uint256[5] _subscriptionParams) internal {
-        require(_subscriptionParams[0] < MAX_UINT64, ERROR_OVERFLOW); // _periodDuration
-        require(_subscriptionParams[3] < MAX_UINT16, ERROR_OVERFLOW); // _latePaymentPenaltyPct
-        require(_subscriptionParams[4] < MAX_UINT16, ERROR_OVERFLOW); // _governorSharePct
+    function _initSubscriptions(ERC20 _feeToken, uint256[5] _subscriptionParams, ISumTree _sumTree) internal {
+        require(_subscriptionParams[0] <= MAX_UINT64, ERROR_OVERFLOW); // _periodDuration
+        require(_subscriptionParams[3] <= MAX_UINT16, ERROR_OVERFLOW); // _latePaymentPenaltyPct
+        require(_subscriptionParams[4] <= MAX_UINT16, ERROR_OVERFLOW); // _governorSharePct
         subscriptions.init(
             ISubscriptionsOwner(this),
-            sumTree,
+            _sumTree,
             uint64(_subscriptionParams[0]), // _periodDuration
             _feeToken,
             _subscriptionParams[1],         // _feeAmount
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -6,6 +6,8 @@ import "./standards/arbitration/IArbitrable.sol";
 import "./standards/erc900/ERC900.sol";
 import "./standards/voting/ICRVoting.sol";
 import "./standards/voting/ICRVotingOwner.sol";
+import "./standards/subscription/ISubscription.sol";
+import "./standards/subscription/ISubscriptionOwner.sol";
 
 import { ApproveAndCallFallBack } from "@aragon/apps-shared-minime/contracts/MiniMeToken.sol";
 import "@aragon/os/contracts/lib/token/ERC20.sol";
@@ -14,7 +16,7 @@ import "@aragon/os/contracts/lib/math/SafeMath.sol";
 
 
 // solium-disable function-order
-contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
+contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptionOwner {
     using SafeERC20 for ERC20;
     using SafeMath for uint256;
 
@@ -105,14 +107,14 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
     ISumTree internal sumTree;
 
     // Global config, configurable by governor
-    address public governor; // TODO: consider using aOS' ACL
+    address internal governor; // TODO: consider using aOS' ACL
     uint256 public jurorMinStake; // TODO: consider adding it to the conf
     CourtConfig[] public courtConfigs;
 
     // Court state
     uint64 public termId;
     uint64 public configChangeTermId;
-    mapping (address => Account) public accounts;
+    mapping (address => Account) internal accounts;
     mapping (uint256 => address) public jurorsByTreeId;
     mapping (uint64 => Term) public terms;
     Dispute[] public disputes;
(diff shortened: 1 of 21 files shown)


contracts/Court.sol
```diff
@@ -426,12 +428,9 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
 
         require(dispute.state == DisputeState.PreDraft, ERROR_ROUND_ALREADY_DRAFTED);
-        require(_blockNumber() > draftTerm.randomnessBN, ERROR_TERM_RANDOMNESS_UNAVAIL);
         require(round.draftTermId <= termId, ERROR_NOT_DRAFT_TERM);
-
-        if (draftTerm.randomness == bytes32(0)) {
-            draftTerm.randomness = blockhash(draftTerm.randomnessBN);
-        }
+        // Ensure that term has randomness:
+        _getTermRandomness(draftTerm);
```
bingen: Should we change its name? As it's not a view function and this one suggest it's just a getter.

contracts/Court.sol
```diff
@@ -851,6 +850,35 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         return disputes[_disputeId].rounds[_roundId].jurorSlotStates[_juror].weight;
     }
 
+    /* Subscription interface */
+    function getCurrentTermId() external ensureTerm returns (uint64) {
```
bingen: See comment above about the name.

contracts/Subscription.sol
```diff
@@ -0,0 +1,237 @@
...
+    string internal constant ERROR_TOKEN_TRANSFER_FAILED = "SUB_TOKEN_TRANSFER_FAILED";
+    string internal constant ERROR_ZERO_PERIOD_DURATION = "SUB_ZERO_PERIOD_DURATION";
+    string internal constant ERROR_ZERO_FEE = "SUB_ZERO_FEE";
+    string internal constant ERROR_ZERO_PREPAYMENT_PERIODS = "SUB_ZERO_PREPAYMENT_PERIODS";
+    string internal constant ERROR_INVALID_PERIOD = "SUB_INVALID_PERIOD";
+    string internal constant ERROR_PAY_ZERO_PERIODS = "SUB_PAY_ZERO_PERIODS";
+    string internal constant ERROR_TOO_MANY_PERIODS = "SUB_TOO_MANY_PERIODS";
+
+    struct Organization {
+        uint256 lastPaymentPeriodId;
```
bingen: We could get rid of the struct.

contracts/Subscription.sol
```diff
@@ -0,0 +1,237 @@
...
+                randomness = blockhash(block.number - 1);
+            }
+            periodBalanceCheckpoint = periodStartTermId + uint64(uint256(randomness) % periodDuration);
+            period.balanceCheckpoint = periodBalanceCheckpoint;
+        }
+
+        // get balance and total at checkpoint
+        uint256 sumTreeId = owner.getAccountSumTreeId(msg.sender);
+        uint256 jurorBalance = sumTree.getItemPast(sumTreeId, periodBalanceCheckpoint);
+        uint256 totalTreeSum = period.totalTreeSum;
```
bingen: This is not being set anywhere!

contracts/Subscription.sol
```diff
@@ -0,0 +1,237 @@
...
+            // if there are more pending payments than requested to pay, adjust them
+            if (delayedPeriods > _periods) {
+                delayedPeriods = _periods;
+            } else { // otherwise the rest are regular payments
+                regularPeriods = _periods - delayedPeriods;
+            }
+        }
+
+        // don't allow to pay too many periods in advance (see comments in declaration section)
+        require(regularPeriods <= prePaymentPeriods, ERROR_TOO_MANY_PERIODS);
```
bingen: We should check `lastPaymentPeriodId` here

contracts/Subscription.sol
```diff
@@ -0,0 +1,237 @@
...
+
+        // total amount to pay by sender (on behalf of org), including penalties for delayed periods
+        uint256 amountToPay = _pct4_increase(delayedPeriods.mul(feeAmount), latePaymentPenaltyPct).add(regularPeriods.mul(feeAmount));
+        // governor fee
+        uint256 governorFee = _pct4(amountToPay, governorSharePct);
+        accumulatedGovernorFees += governorFee;
+        // amount collected for the current period to share among jurors
+        uint256 collectedFees = amountToPay - governorFee;
+
+        organization.lastPaymentPeriodId = currentPeriodId + regularPeriods;
```
bingen: Here `currentPeriodId` should be `lastPaymentPeriodId`

contracts/Subscription.sol
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    // getCurrentTermId in ISubscriptionOwner is not view, because of ensureTerm
+    function _getCurrentPeriodId() internal returns (uint256) {
+        // periodId 0 is reserved to signal non-subscribed organizations
+        return (owner.getCurrentTermId() - startTermId) / periodDuration + 1;
+    }
+
+    function _getPeriodStartTermId(uint256 _periodId) internal view returns (uint64) {
+        return startTermId + uint64(_periodId) * periodDuration;
```
bingen: Missing adjustment for non-used periodId 0

contracts/Subscription.sol
```diff
@@ -0,0 +1,237 @@
...
+            if (randomness == bytes32(0)) {
+                randomness = blockhash(block.number - 1);
+            }
+            periodBalanceCheckpoint = periodStartTermId + uint64(uint256(randomness) % periodDuration);
+            period.balanceCheckpoint = periodBalanceCheckpoint;
+        }
+
+        // get balance and total at checkpoint
+        uint256 sumTreeId = owner.getAccountSumTreeId(msg.sender);
+        uint256 jurorBalance = sumTree.getItemPast(sumTreeId, periodBalanceCheckpoint);
```
bingen: We must check it's not zero.

contracts/Court.sol
izqui: Rather than performing the term transition here we could just calculate how many transitions would be needed (`neededTermTransitions()`) and add it to the current `termId`. For the purpose of subscriptions we only care about what the term number is and not whether all transitions have been done.

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+        uint256 amount = accumulatedGovernorFees;
+        accumulatedGovernorFees = 0;
+        require(feeToken.safeTransfer(owner.getGovernor(), amount), ERROR_TOKEN_TRANSFER_FAILED);
+    }
+
+    function setFeeAmount(uint256 _feeAmount) external onlyOwner {
+        _setFeeAmount(_feeAmount);
+    }
+
+    function setFeeToken(ERC20 _feeToken) external onlyOwner {
```
izqui: 🚨 Changing the fee token would be catastrophic as we only keep track of fee amounts. A simple workaround for when the fee token is changed could be:

- The first time fees are paid for a period, cache the `feeToken` in the period struct and have all payments in the period use the cached `feeToken` even if it has been modified.
- Execute `transferFeesToGovernor` so we clean up `accumulatedGovernorFees`

Edit: This is also problematic for the fee amount, as the token address and the fee amount are quite related (unless the tokens have the same dollar value, changing the fee token will always require changing the amount). I would merge changing both to one function and cache the fee amount in the period as well (which comes with the nice side effect that everyone will pay the same fee in the period).

test/subscription-fees.js
```diff
@@ -0,0 +1,195 @@
...
+
+        const r = await this.subscription.payFees(org1, 1, { from: org1 })
+
+        const finalBalance = await token.balanceOf(org1)
+
+        assertEqualBNs(initialBalance.sub(FEE_AMOUNT), finalBalance, 'Token balance mismatch');
+        assert.isTrue(await this.subscription.isUpToDate(org1))
+      })
+
+      it('Org fails paying fees too far in the future', async () => {
```
izqui: Would be great to have tests for successfully paying some periods in advance and for an organization that paid fees, then stopped paying for some periods and paying with the penalty.

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+        feeToken = _feeToken;
```
izqui: We probably want to ensure that `feeToken` is a contract

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        // get balance and total at checkpoint
+        uint256 sumTreeId = owner.getAccountSumTreeId(msg.sender);
+        uint256 jurorBalance = sumTree.getItemPast(sumTreeId, periodBalanceCheckpoint);
+        require(jurorBalance > 0, ERROR_NOTHING_TO_CLAIM);
+        // it can't happen that total sum is zero if juror's balance is not
+
+        // juror fee share
+        uint256 jurorShare = period.collectedFees.mul(jurorBalance) / period.totalTreeSum;
```
izqui: Similar to the comment in `payFees`, it would be great to have a getter for the reward that a juror will get so they can decide whether to claim the fees or not. Right now the frontend would need to re-implement all the logic client-side (and some information would be really hard to get)

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+    uint16 public governorSharePct; // ‱ of fees that go to governor of the Court
+    ERC20 public feeToken;
+    // How many periods can be paid in advance (includes current period, so it must be at least 1).
+    // Although paying in advance seems a good thing from the Court perspective,
+    // it has some drawbacks too, so it's good to limit it to diminish them:
+    // - Fees distribution among jurors take place when the payment is made, so jurors activating after a pre-payment wouldn't get their share of it.
+    // - Fee amount could increase, while pre-payments would be made with the old rate.
+    uint256 public prePaymentPeriods;
+    uint256 public feeAmount;
+    uint256 internal accumulatedGovernorFees;
```
izqui: Why make this internal?

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+        uint256 lastPaymentPeriodId;
+    }
+
+    struct Period {
+        uint64 balanceCheckpoint;
+        uint256 totalTreeSum;
+        uint256 collectedFees;
+    }
+
+    ISubscriptionOwner internal owner;
```
izqui: Why make this internal?

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+        uint16 _governorSharePct
+    )
+        external
+    {
+        require(address(owner) == address(0), ERROR_OWNER_ALREADY_SET);
+        require(_periodDuration > 0, ERROR_ZERO_PERIOD_DURATION);
+
+        owner = _owner;
+        sumTree = _sumTree;
+        startTermId = _startTermId; // Court should set this to > 0, as term 0 is for jurors onboarding
```
izqui: Any reason to make it non-configurable then? It could actually even be a constant

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+    string internal constant ERROR_TOKEN_TRANSFER_FAILED = "SUB_TOKEN_TRANSFER_FAILED";
+    string internal constant ERROR_ZERO_PERIOD_DURATION = "SUB_ZERO_PERIOD_DURATION";
+    string internal constant ERROR_ZERO_FEE = "SUB_ZERO_FEE";
+    string internal constant ERROR_ZERO_PREPAYMENT_PERIODS = "SUB_ZERO_PREPAYMENT_PERIODS";
+    string internal constant ERROR_INVALID_PERIOD = "SUB_INVALID_PERIOD";
+    string internal constant ERROR_NOTHING_TO_CLAIM = "SUB_NOTHING_TO_CLAIM";
+    string internal constant ERROR_PAY_ZERO_PERIODS = "SUB_PAY_ZERO_PERIODS";
+    string internal constant ERROR_TOO_MANY_PERIODS = "SUB_TOO_MANY_PERIODS";
+
+    struct Organization {
```
izqui: I would rename to `Subscriber` everywhere to make it more agnostic

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+import "./standards/subscription/ISubscription.sol";
+import "./standards/subscription/ISubscriptionOwner.sol";
+import "./standards/sumtree/ISumTree.sol";
+
+import "@aragon/os/contracts/lib/token/ERC20.sol";
+import "@aragon/os/contracts/common/SafeERC20.sol";
+import "@aragon/os/contracts/lib/math/SafeMath.sol";
+
+
+contract Subscription is ISubscription {
```
izqui: I would rename it to `CourtSubscriptions` or `Subscriptions`

contracts/Court.sol
```diff
@@ -851,6 +850,35 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
...
+        return termId;
+    }
+
+    function getTermRandomness(uint64 _termId) external returns (bytes32) {
```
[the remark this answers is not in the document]
bingen: Good question. I would say strictly speaking it doesn't, as you could be asking for a previous term. But then at the very least needs `require(_termId <= termId);`.
Anyway, as the function isn't `view` already, I'll add the `ensureTerm` too.

contracts/Subscription.sol
```diff
@@ -0,0 +1,244 @@
...
+                delayedPeriods = _periods;
+                // regular periods will therefore be zero, as all to be paid periods are past ones
+            } else { // otherwise the rest are regular payments
+                regularPeriods = _periods - delayedPeriods;
+            }
+        }
+
+        // don't allow to pay too many periods in advance (see comments in declaration section)
+        uint256 newLastPeriodId = nextPaymentPeriodId + regularPeriods - 1;
+        require(newLastPeriodId <= prePaymentPeriods, ERROR_TOO_MANY_PERIODS);
```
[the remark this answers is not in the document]
bingen: Yes, this is wrong, but it should be `newLastPeriodId - currentPeriodId < prePaymentPeriods`. This way an org can not pre-pay too far in the future by calling the function repeatedly.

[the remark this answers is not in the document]
bingen: Btw, I'm afraid previous line was wrong too.

contracts/Subscription.sol
```diff
@@ -0,0 +1,244 @@
...
+        // periodId 0 is reserved to signal non-subscribed organizations
+        require(_periodId > 0, ERROR_INVALID_PERIOD);
+        return startTermId + uint64(_periodId - 1) * periodDuration;
+    }
+
+    function _pct4(uint256 _number, uint16 _pct) internal pure returns (uint256) {
+        return _number * uint256(_pct) / uint256(PCT_BASE);
+    }
+
+    function _pct4_increase(uint256 _number, uint16 _pct) internal pure returns (uint256) {
```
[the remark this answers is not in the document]
bingen: Yes, linter told me this one too ;-)

contracts/Subscription.sol
bingen: I was using `owner()` instead of the newly created `getOwner()` and was having some weird issues running tests. I think it was that one that says not enough gas when an interface is not met.

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+                regularPeriods = _periods - delayedPeriods;
+            }
+        }
+
+        // don't allow to pay too many periods in advance (see comments in declaration section)
+        uint256 newLastPeriodId = nextPaymentPeriodId + regularPeriods - 1;
+        require(newLastPeriodId <= prePaymentPeriods, ERROR_TOO_MANY_PERIODS);
+
+        // total amount to pay by sender (on behalf of org), including penalties for delayed periods
+        uint256 amountToPay = _pct4_increase(delayedPeriods.mul(feeAmount), latePaymentPenaltyPct).add(regularPeriods.mul(feeAmount));
```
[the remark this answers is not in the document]
bingen: But you mean passing `_periods` to that getter? Or only for the delayed ones?

contracts/Subscription.sol
```diff
@@ -0,0 +1,244 @@
...
+        require(feeToken.safeTransfer(msg.sender, jurorShare), ERROR_TOKEN_TRANSFER_FAILED);
+    }
+
+    function transferFeesToGovernor() external {
+        uint256 amount = accumulatedGovernorFees;
+        accumulatedGovernorFees = 0;
+        require(feeToken.safeTransfer(owner.getGovernor(), amount), ERROR_TOKEN_TRANSFER_FAILED);
+    }
+
+    function setFeeAmount(uint256 _feeAmount) external onlyOwner {
```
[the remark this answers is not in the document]
bingen: Yes, I like the idea, specially because Court size is so close to the limit. My only concern is that if `governor` is a contract, it has to be able to call them (like a DAO having an agent, or whatever). Should we add an `IGovernor` interface?

contracts/Subscription.sol
```diff
@@ -0,0 +1,244 @@
...
+
+    function setFeeToken(ERC20 _feeToken) external onlyOwner {
+        feeToken = _feeToken;
+    }
+
+    function setPrePaymentPeriods(uint256 _prePaymentPeriods) external onlyOwner {
+        _setPrePaymentPeriods(_prePaymentPeriods);
+    }
+
+    function setLatePaymentPenaltyPct(uint16 _latePaymentPenaltyPct) external onlyOwner {
```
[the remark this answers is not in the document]
bingen: Hm, not so sure about this. I already fixed that one and 100%, although reasonable, sounds a little bit arbitrary. What if a Court wants to be aggressive with due payments and do a 3x penalty? Actually there's already a natural limit of ~7.5x due to it being `uint16`, so I don't see the need for adding more code.

contracts/Subscription.sol
```diff
@@ -0,0 +1,242 @@
...
+        // periodId 0 is reserved to signal non-subscribed organizations
+        if (nextPaymentPeriodId == 1) {
+            // not yet subscribed orgs can't have pending payments
+            regularPeriods = _periods;
+        } else {
+            // check for pending payments
+            if (currentPeriodId > nextPaymentPeriodId) {
+                delayedPeriods = currentPeriodId - nextPaymentPeriodId;
+            }
+            // if there are more pending payments than requested to pay, adjust them
```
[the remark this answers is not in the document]
bingen: Exactly, it wouldn't catch up completely. It could go, for instance, from 5 overdue period payments to only 2, but it wouldn't be up to date yet.

contracts/Subscription.sol
bingen: What do you mean with "to make it non-configurable" ? That you would remove it from the constructor?

izqui: Yes, and just assume it is 1.

contracts/Subscription.sol
```diff
@@ -84,56 +89,47 @@ contract Subscription is ISubscription {
         require(_periodDuration > 0, ERROR_ZERO_PERIOD_DURATION);
 
         owner = _owner;
+        governor = owner.getGovernor(); // This is to avoid making an external call in onlyGovernor modifier
```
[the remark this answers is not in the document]
bingen: Yes, that was my initial idea, but then I realized that governor can't currently be changed in Court (are we planning to do it?).
Besides it's supposed not to be a [good practice](https://consensys.github.io/smart-contract-best-practices/recommendations/#use-modifiers-only-for-assertions), but I guess we can trust owner (aka Court) in this case, as they are tightly coupled.
So I'll revert this.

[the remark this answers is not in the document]
izqui: Yes, we should be able to modify the governor in the Court. We can totally trust the Court not to re-enter here IMO.

contracts/Subscription.sol
```diff
@@ -142,44 +138,23 @@ contract Subscription is ISubscription {
...
+    function isUpToDate(address _subscriber) external view returns (bool) {
+        Subscriber storage subscriber = subscribers[_subscriber];
+        return subscriber.subscribed && subscriber.lastPaymentPeriodId >= _getCurrentPeriodId();
     }
 
     function claimFees(uint256 _periodId) external {
-        // periodId 0 is reserved to signal non-subscribed organizations
-        require(_periodId > 0 && _periodId < _getCurrentPeriodId(), ERROR_INVALID_PERIOD);
+        require(_periodId < _getCurrentPeriodId(), ERROR_INVALID_PERIOD);
+        require(!claimedFees[msg.sender][_periodId], ERROR_ALREADY_CLAIMED);
```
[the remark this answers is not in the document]
bingen: Hm, but what's the advantage? The way it is right now:
- The message will be more accurate (It will fail sooner if a juror claims twice, so less gas).
- The getter function will always return the same, and will be honoring its name better: I think the juror share is the same regardless it has been claimed or not. (Maybe we can add a getter to check if a juror has claimed).

contracts/Subscription.sol
```diff
@@ -0,0 +1,343 @@
...
+
+    function setLatePaymentPenaltyPct(uint16 _latePaymentPenaltyPct) external onlyGovernor {
+        latePaymentPenaltyPct = _latePaymentPenaltyPct;
+    }
+
+    function setGovernorSharePct(uint16 _governorSharePct) external onlyGovernor {
+        governorSharePct = _governorSharePct;
+    }
+
+    function ensurePeriodBalanceCheckpoint(uint256 _periodId) external returns (uint64) {
```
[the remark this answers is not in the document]
bingen: Well to be honest I added it by inertia and then I thought that it may be useful if for instance the governor wants to call it at the beginning of each period so the first subscriber paying fees is not penalized paying more gas for being the first one (and thus to avoid incentivizing paying at the end of the period). But I guess it's overthinking. :sweat_smile: 
What you say is even more interesting, I didn't think of it, but now I'm leaning towards leaving it.

contracts/Subscription.sol
```diff
@@ -0,0 +1,343 @@
...
+        uint256 sumTreeId = owner.getAccountSumTreeId(_juror);
+        uint256 jurorBalance = sumTree.getItemPast(sumTreeId, _periodBalanceCheckpoint);
+
+        if (jurorBalance == 0) {
+            return 0;
+        }
+        // it can't happen that total sum is zero if juror's balance is not
+
+        // juror fee share
+        return _period.collectedFees.mul(jurorBalance) / _period.totalTreeSum;
```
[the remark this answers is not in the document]
bingen: Good catch!

contracts/Subscription.sol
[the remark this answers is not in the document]
izqui: Good point. Let's definitely add a getter for whether the juror already claimed their fees then

contracts/Subscription.sol
bingen: The problem with using 0 as a valid period is that if org is not subscribed yet, original lastPaymentPeriodId should be -1, but it will be 0 instead. So we have some ugliness here too. But anyway I'm keeping your version.

contracts/CourtSubscriptions.sol
```diff
@@ -0,0 +1,475 @@
...
+        Period storage period = periods[currentPeriodId];
+
+        (ERC20 feeToken, uint256 feeAmount) = _ensurePeriodFeeTokenAndAmount(period);
+
+        // total amount to pay by sender (on behalf of org), including penalties for delayed periods
+        (uint256 amountToPay, uint256 newLastPeriodId) = _getPayFeesDetails(subscriber, _periods, currentPeriodId, feeAmount);
+        // as _periods and feeAmount are > 0, amountToPay will be > 0 and newLastPeriod > subscriber.lastPaymentPeriodId
+        // governor fee
+        uint256 governorFee = _pct4(amountToPay, governorSharePct);
+        accumulatedGovernorFees += governorFee;
```
izqui: Can we move this closer to the line where we set `period.collectedFees += collectedFees;`

contracts/CourtSubscriptions.sol
```diff
@@ -0,0 +1,475 @@
...
+        Period storage _period,
+        uint64 _periodBalanceCheckpoint,
+        uint256 _totalTreeSum
+    )
+        internal
+        view
+        returns (uint256)
+    {
+        // get balance and total at checkpoint
+        uint256 sumTreeId = owner.getAccountSumTreeId(_juror);
```
izqui: If the `sumTreeId` is 0, we could return 0 immediately, as it means that the account wasn't activated as a juror

contracts/CourtSubscriptions.sol
```diff
@@ -0,0 +1,475 @@
...
+        uint256 jurorBalance = sumTree.getItemPast(sumTreeId, _periodBalanceCheckpoint);
+
+        if (jurorBalance == 0) {
+            return 0;
+        }
+        // it can't happen that total sum is zero if juror's balance is not
```
izqui:
```suggestion
        // Invariant: If the jurorBalance is greater than 0, the totalSum must be greater than 0 as well
```

contracts/CourtSubscriptions.sol
```diff
@@ -0,0 +1,475 @@
...
+        uint64 periodStartTermId = _getPeriodStartTermId(_periodId);
+        uint64 nextPeriodStartTermId = _getPeriodStartTermId(_periodId + 1);
+        bytes32 randomness = owner.getTermRandomness(nextPeriodStartTermId);
+        // if randomness was not calculated on first 256 blocks of the term, it will be zero
+        // in that case we just get the previous block for it, as we'll have the hash for sure
+        // it could be slightly beneficial for the first juror calling this function,
+        // but it's still impossible to predict during the period
+        if (randomness == bytes32(0)) {
+            randomness = blockhash(block.number - 1);
+        }
```
izqui: A comment along these lines would be great: "A Court term during the previous period is selected randomly and it is used as the checkpoint for distributing the fees collected during the period"

contracts/CourtSubscriptions.sol
```diff
@@ -0,0 +1,475 @@
...
+            if (delayedPeriods > _periods) {
+                delayedPeriods = _periods;
+                // regular periods will therefore be zero, as all to be paid periods are past ones
+            } else { // otherwise the rest are regular payments
+                regularPeriods = _periods - delayedPeriods;
+            }
+        }
+
+        // don't allow to pay too many periods in advance (see comments in declaration section)
+        newLastPeriodId = lastPaymentPeriodId + delayedPeriods + regularPeriods;
```
izqui: As discussed offline, we should take into account that when someone subscribes after the first period, they shouldn't have to pay fees for all the past periods.

contracts/Court.sol
```diff
@@ -389,8 +401,9 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner {
         // TODO: Limit the max amount of terms into the future that a dispute can be drafted
         // TODO: Limit the max number of initial jurors
         // TODO: ERC165 check that _subject conforms to the Arbitrable interface
-        // TODO: Consider requiring that only the contract being arbitred can create a dispute
 
+        // TODO: require(address(_subject) == msg.sender, ERROR_INVALID_DISPUTE_CREATOR);
```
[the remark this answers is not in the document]
bingen: Because I realized it would break a lot of tests and it actually doesn't belong in here. (I shouldn't have touched it)

bingen:
Closed in favor of #63 (keeping this one for discussion reference)

PR #58, opened 2019-07-03
Court: prevents drafting without randomness seed

### LINKED ISSUE #15: Prevent drafts happening without a randomness seed
Right now, if the first draft of a term happens more than 256 blocks after the term starts, the blockhash that will be returned for the `randomnessBN` of the term will be 0.

Options to work around this problem:

- Treat this as a round being drafted in the wrong term and require rescheduling as in #14 
- Implement a secondary `heartbeat()` function (which can be called one block after the actual `heartbeat`) that stores the `randomness` for the term. The caller could get the current `dependingDrafts * heartbeatFee` that is being paid to the `heartbeat()` caller and pay heartbeat fees just from activation, updates and deactivation fees (#7)

bingen:
Drafting will have to wait for a later term where randomness seed is
available.

Closes #15.

### DIFF
--- a/test/court-batches.js
+++ b/test/court-batches.js
@@ -54,6 +54,8 @@ contract('Court: Batches', ([ rich, governor, arbitrable, juror1, juror2, juror3
   const JUROR_DRAFTED_EVENT = 'JurorDrafted'
   const DISPUTE_STATE_CHANGED_EVENT = 'DisputeStateChanged'
 
+  const ERROR_TERM_RANDOMNESS_UNAVAIL = 'COURT_TERM_RANDOMNESS_UNAVAIL'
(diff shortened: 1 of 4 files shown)


test/court-batches.js
```diff
@@ -237,6 +241,23 @@ contract('Court: Batches', ([ rich, governor, arbitrable, juror1, juror2, juror3
         assert.isTrue(termsPassed > 1, 'draft was not split')
       })
 
+
+      it('needs to wait until next term if randomness is missing', async () => {
+        let totalJurorsDrafted = 0
```
izqui: We can clean up a bunch of unused variables in this test

PR #59, opened 2019-07-08
Court: Make max number of appeal rounds configurable

### LINKED ISSUE #55: Make max number of appeal rounds configurable
See: https://github.com/aragon/aragon-court/pull/45#discussion_r287787147

bingen:
Fixes #55.

:warning: It increments bytecode by ~800, surpassing EIP-170 value again.

### DIFF
--- a/contracts/Court.sol
+++ b/contracts/Court.sol
@@ -21,7 +21,6 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
     using SafeMath for uint256;
 
     uint256 internal constant MAX_JURORS_PER_DRAFT_BATCH = 10; // to cap gas used on draft
-    uint256 internal constant MAX_REGULAR_APPEAL_ROUNDS = 4; // before the final appeal
     uint256 internal constant FINAL_ROUND_WEIGHT_PRECISION = 1000; // to improve roundings
     uint64 internal constant APPEAL_STEP_FACTOR = 3;
     // TODO: move all other constants up here
@@ -90,6 +89,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
         IArbitrable subject;
         uint8 possibleRulings;      // number of possible rulings the court can decide on
         uint8 winningRuling;
+        uint32 maxRegularAppealRounds; // before the final appeal
         DisputeState state;
         AdjudicationRound[] rounds;
     }
@@ -108,6 +108,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
     // Court state
     uint64 internal termId;
     uint64 public configChangeTermId;
+    uint32 public maxRegularAppealRounds; // before the final appeal
     mapping (uint64 => Term) public terms;
     Dispute[] public disputes;
 
@@ -136,6 +137,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
     string internal constant ERROR_JUROR_ALREADY_REWARDED = "CTJUROR_ALRDY_REWARDED";
     string internal constant ERROR_JUROR_NOT_COHERENT = "CTJUROR_INCOHERENT";
     string internal constant ERROR_WRONG_PENALTY_PCT = "CTBAD_PENALTY";
+    string internal constant ERROR_ZERO_MAX_ROUNDS = "COURT_ZERO_MAX_ROUNDS";
 
     uint64 internal constant ZERO_TERM_ID = 0; // invalid term that doesn't accept disputes
     uint64 internal constant MODIFIER_ALLOWED_TERM_TRANSITIONS = 1;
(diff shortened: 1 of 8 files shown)


contracts/Court.sol
```diff
@@ -725,6 +730,11 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
         feeAmount = fees.heartbeatFee + _jurorNumber * (fees.jurorFee + fees.draftFee + fees.settleFee);
     }
 
+    function setMaxRegularAppealRounds(uint32 _maxRegularAppealRounds) public {
```
facuspagnuolo: we should restrict the access of this function only to the owner or the governor

contracts/Court.sol
```diff
@@ -96,6 +95,7 @@ contract Court is ERC900, ApproveAndCallFallBack, ICRVotingOwner, ISubscriptions
         IArbitrable subject;
         uint8 possibleRulings;      // number of possible rulings the court can decide on
         uint8 winningRuling;
+        uint32 maxRegularAppealRounds; // before the final appeal
```
izqui: Wondering whether we should make it part of the global Court configuration instead of caching it in every dispute. The `maxRegularAppealRounds` for a dispute would be gotten from the Court config at the term of its first rounds' draft. Thoughts @bingen @facuspagnuolo?

facuspagnuolo: The only thing will be that it can be changed during a dispute itself since the court config is being fetched using the term id of each round... which may be weird for other settings as well now that I'm thinking of...

facuspagnuolo: I already moved the max regular appeals number to the court settings and I will create another issue to think a bit more about the way we are handling the updates for the court settings

contracts/Court.sol
```diff
@@ -395,18 +398,19 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
         Dispute storage dispute = disputes[_disputeId];
         AdjudicationRound storage currentRound = dispute.rounds[_roundId];
+        CourtConfig storage config = courtConfigs[terms[currentRound.draftTermId].courtConfigId]; // safe to use directly as it is the current term
 
         uint64 appealJurorNumber;
         uint64 appealDraftTermId = termId + 1; // Appeals are drafted in the next term
 
         uint256 roundId;
-        if (_roundId == MAX_REGULAR_APPEAL_ROUNDS - 1) { // final round, roundId starts at 0
+        if (_roundId == config.maxRegularAppealRounds - 1) { // final round, roundId starts at 0
```
izqui:
```suggestion
        if (_roundId >= config.maxRegularAppealRounds - 1) { // final round, roundId starts at 0
```

Given that we are using the config for the draft term for the round to decide whether or not to create a regular appeal, it is possible that `_roundId` is never `config.maxRegularAppealRounds - 1` because the `maxRegularAppealRounds` could be decreased.

If for the initial rounds `maxRegularAppealRounds = 4`, and during  `_roundId = 2` it is decreased to `maxRegularAppealRounds = 2`, when appealing it will create a regular appeal round (because at the draft term of the round the max was still 4). But when _appealing `roundId = 3`, instead of starting the final round (as it should), it will go to the else statement and fail the assertion in L413.

I think for now we should just start the final round in case this happens, but we should definitely revisit it with #90

contracts/Court.sol
```diff
@@ -660,8 +664,12 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
     function canCommit(uint256 _voteId, address _voter) external ensureTerm only(voting) returns (uint256 weight) {
         (uint256 disputeId, uint256 roundId) = _decodeVoteId(_voteId);
 
+        Dispute storage dispute = disputes[disputeId];
+        AdjudicationRound storage round = dispute.rounds[roundId];
+        CourtConfig storage config = courtConfigs[terms[round.draftTermId].courtConfigId];
+
         // for the final round
-        if (roundId == MAX_REGULAR_APPEAL_ROUNDS) {
+        if (roundId == config.maxRegularAppealRounds) {
```
izqui:
```suggestion
        if (roundId >= config.maxRegularAppealRounds) {
```

contracts/Court.sol
```diff
@@ -910,6 +920,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
         // We make sure that when applying penalty pct to juror min stake it doesn't result in zero
         uint256 minJurorsActiveBalance = jurorsRegistry.minJurorsActiveBalance();
         require(uint256(_penaltyPct) * minJurorsActiveBalance >= PCT_BASE, ERROR_WRONG_PENALTY_PCT);
+        require(_maxRegularAppealRounds > uint32(0), ERROR_ZERO_MAX_ROUNDS);
```
izqui: Should we hardcode a maximum value as well? 10 sounds like a reasonable limit

contracts/Court.sol
```diff
@@ -21,7 +21,6 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
     using SafeMath for uint256;
 
     uint256 internal constant MAX_JURORS_PER_DRAFT_BATCH = 10; // to cap gas used on draft
-    uint256 internal constant MAX_REGULAR_APPEAL_ROUNDS = 4; // before the final appeal
     uint256 internal constant FINAL_ROUND_WEIGHT_PRECISION = 1000; // to improve roundings
     uint64 internal constant APPEAL_STEP_FACTOR = 3;
```
izqui: It probably makes sense to allow configuring this value as well.

PR #65, opened 2019-07-17
Split final round

bingen:


izqui:
As discussed offline, we are deprioritizing splitting the final round logic from the main Court contract. We should close for now and perhaps work on it later if we decide to do a bigger re-architecture (e.g. abstracting adjudication mechanisms separating the logic of regular rounds as well)

PR #67, opened 2019-07-25
Court accounting

facuspagnuolo:
This PR introduces the concept of `CourtAccounting`, a new contract responsible for holding funds related to fees and handling withdraws to the corresponding holders.


contracts/CourtAccounting.sol
```diff
@@ -0,0 +1,59 @@
...
+        require(msg.sender == owner, ERROR_SENDER_NOT_OWNER);
+        _;
+    }
+
+    function init(address _owner) external {
+        require(owner == address(0), ERROR_ALREADY_INITIALIZED);
+        owner = _owner;
+    }
+
+    function deposit(ERC20 _token, address _to, uint256 _amount) external onlyOwner {
```
izqui: I think this function name is a bit strange given that the actual deposit of tokens in this contract must have already happened. I would rename it to `assign` or `debit` as what we are doing is just allowing the receiver to withdraw tokens in the future.

contracts/Court.sol
```diff
@@ -802,7 +808,8 @@ contract Court is IStakingOwner, ICRVotingOwner, ISubscriptionsOwner {
         terms[_draftTermId].dependingDrafts += 1;
 
         if (_feeAmount > 0) {
-            require(_feeToken.safeTransferFrom(msg.sender, address(staking), _feeAmount), ERROR_DEPOSIT_FAILED);
+            // TODO: can we do the transfer-from from the accounting itself? the disputer must approve the accounting contract in that case
```
izqui: I think it is cleaner to approve the same contract that you interact with as it is right now, but either way is fine with me.

PR #69, opened 2019-07-30
Bonding curve

bingen:
NOTICE: Change base to `development` once #64 has been merged.

Integrate bonding curve from Fundraising project
    
Just import it and add a simple test file to check that the whole flow works: buying into the curve and then staking and activating.

TODO:

- [ ] Update all tests to Truffle 5

### DIFF
--- a/contracts/test/TestImports.sol
+++ b/contracts/test/TestImports.sol
@@ -1,6 +1,14 @@
 pragma solidity ^0.4.24;
 
 import "@aragon/apps-shared-migrations/contracts/Migrations.sol";
+import "@aragon/apps-shared-minime/contracts/MiniMeToken.sol";
+
+import "@aragon/os/contracts/kernel/Kernel.sol";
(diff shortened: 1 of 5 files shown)


bingen:
> IMO if we are testing the bonding curve isolated doesn't make much sense, they should be doing that...

Yes, I agree. There's already a good set of tests in fundraising repo.

> but looking at the tests I think you're using the bonding curve to get some ANJ only

Yes that was the idea, and what we discussed offline. Besides a small sanity check [here](https://github.com/aragon/aragon-court/pull/69/files#diff-5f874ee5625e6e887238ddaf1c6144a6R253), the only thing being tested is that we can use Fundraising implementation as it is in order for jurors to convert ANT to Court stake through a bonding curve. This I think it makes total sense.
If the question is where to put this test, here or anywhere else, I don't mind. Let me know where you want it and I'll move it.

PR #75, opened 2019-08-01
Appeal collateral

### LINKED ISSUE #91: Implement appeal collaterals
For an appeal to occur, both parties of the appeal must offer collateral. If the current ruling for a dispute was `X` and one of the parties adds collateral in favor of another ruling `Y` but the other party does not back `X` with collateral, the final ruling will be `Y`. If both parties back their appeal with collateral, a new dispute round gets created.

bingen:
Fixes #91


contracts/Court.sol
```diff
@@ -393,23 +421,73 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
         }
     }
 
+    function _endTermForAdjudicationRound(AdjudicationRound storage round) internal view returns (uint64) {
+        uint64 draftTermId = round.draftTermId;
+        uint64 configId = terms[draftTermId].courtConfigId;
+        CourtConfig storage config = courtConfigs[uint256(configId)];
+
+        // TODO: delayed??
```
facuspagnuolo: ?

contracts/Court.sol
```diff
@@ -393,23 +421,73 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+        return draftTermId + config.commitTerms + config.revealTerms + config.appealTerms + config.appealConfirmTerms;
+    }
+
     /**
      * @notice Appeal round #`_roundId` ruling in dispute #`_disputeId`
      */
-    function appealRuling(uint256 _disputeId, uint256 _roundId) external ensureTerm {
-        _checkAdjudicationState(_disputeId, _roundId, AdjudicationState.Appealable);
+    function appealRuling(uint256 _disputeId, uint256 _roundId, uint8 forRuling) external ensureTerm {
```
facuspagnuolo: Use `_ruling` instead of `forRuling`

contracts/Court.sol
```diff
@@ -54,7 +58,8 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
         Invalid,
         Commit,
         Reveal,
-        Appealable,
+        Appeal,
+        AppealConfirm,
         Ended
```
facuspagnuolo: We should unify this and use adjectives if we are talking about state

contracts/Court.sol
```diff
@@ -393,23 +421,73 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
         }
     }
 
+    function _endTermForAdjudicationRound(AdjudicationRound storage round) internal view returns (uint64) {
+        uint64 draftTermId = round.draftTermId;
+        uint64 configId = terms[draftTermId].courtConfigId;
+        CourtConfig storage config = courtConfigs[uint256(configId)];
+
+        // TODO: delayed??
+        return draftTermId + config.commitTerms + config.revealTerms + config.appealTerms + config.appealConfirmTerms;
```
facuspagnuolo: This may not be always true for every round right? For example, if it wasn't appealed there is no appeal confirm term I think

contracts/Court.sol
facuspagnuolo: Additionally, I'd call this function simply `appeal`

contracts/Court.sol
```diff
@@ -393,23 +421,73 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+        require(!_isRoundAppealed(currentRound), ERROR_ROUND_ALREADY_APPEALED); // This ruling hasn't been appealed yet
+
+        uint8 currentRuling = _getRoundWinningRuling(_disputeId, _roundId);
+        // check correct ruling
+        require(
+            forRuling > uint8(Ruling.Refused) &&
+            forRuling <= uint8(Ruling.Refused) + MAX_RULING_OPTIONS &&
+            forRuling != currentRuling,
+            ERROR_INVALID_RULING
+        );
```
facuspagnuolo: This validation is awkward, we should improve this and the one in `CRVoting`. We're basically relying in the position of `Refused` in the `Ruling` enum, I think we are messing concepts between a voting state and the actual outcome a juror has voted

contracts/Court.sol
```diff
@@ -776,49 +899,115 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+        appealDeposit = feeAmount * APPEAL_COLLATERAL_FACTOR;
+        appealConfirmDeposit = feeAmount * APPEAL_CONFIRMATION_COLLATERAL_FACTOR;
+    }
+
+    function _getRegularAdjudicationRoundJurorNumber(uint64 _appealStepFactor, uint64 _currentRoundJurorNumber) internal pure returns (uint64 appealJurorNumber) {
+        appealJurorNumber = _appealStepFactor * _currentRoundJurorNumber;
+        // make sure it's odd
+        if (appealJurorNumber % 2 == 0) {
+            appealJurorNumber++;
+        }
```
facuspagnuolo: This logic is repeated above

contracts/Court.sol
```diff
@@ -776,49 +899,115 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+    function _getRoundWinningRuling(uint256 _disputeId, uint256 _roundId) internal view returns (uint8 winningRuling) {
+        uint256 voteId = _getVoteId(_disputeId, _roundId);
+        winningRuling = voting.getWinningRuling(voteId);
 
-        roundId = _createRound(_disputeId, DisputeState.PreDraft, _draftTermId, _jurorNumber, 0, feeToken, feeAmount, jurorFees);
+        // If an appeal was started and not confirmed, the ruling is immediately flipped
+        AdjudicationRound storage round = disputes[_disputeId].rounds[_roundId];
+        if (_isRoundAppealed(round) && !_isRoundAppealConfirmed(round)) {
+            winningRuling = round.appealMaker.forRuling;
+        }
```
facuspagnuolo: Let's use an `if-else` explicitly instead of overwriting a return value

contracts/Court.sol
```diff
@@ -419,11 +497,55 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+            uint8 winningRuling = dispute.winningRuling;
+            uint256 totalDeposit = appealDeposit + appealConfirmDeposit;
+
+            if (appealMaker.forRuling == winningRuling) {
+                accounting.assign(depositToken, appealMaker.appealer, totalDeposit - feeAmount);
+            } else if (appealTaker.forRuling == winningRuling) {
+                accounting.assign(depositToken, appealTaker.appealer, totalDeposit - feeAmount);
+            } else {
+                accounting.assign(depositToken, appealMaker.appealer, appealDeposit - feeAmount / 2);
+                accounting.assign(depositToken, appealTaker.appealer, appealConfirmDeposit - feeAmount / 2);
```
facuspagnuolo: This case should never be reached right?

contracts/Court.sol
```diff
@@ -393,23 +421,73 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
-        uint64 appealDraftTermId = termId + 1; // Appeals are drafted in the next term
+        require(_isRoundAppealed(currentRound), ERROR_ROUND_NOT_APPEALED); // The ruling was appealed
+        require(!_isRoundAppealConfirmed(currentRound), ERROR_ROUND_APPEAL_ALREADY_CONFIRMED); // but not confirmed
+        // check correct ruling
+        require(
+            forRuling > uint8(Ruling.Refused) &&
+            forRuling <= uint8(Ruling.Refused) + MAX_RULING_OPTIONS &&
+            forRuling != currentRound.appealMaker.forRuling,
+            ERROR_INVALID_RULING
+        );
```
facuspagnuolo: Same here

contracts/Court.sol
bingen: Where?

contracts/Court.sol
bingen: Yes, it is, in case of Refused ruling. There even is a test for it.

contracts/Court.sol
bingen: Yes, I forgot this one, thanks for pointing it out! The thing is that a round could be delayed in case not all jurors are drafted in time so there's randomness available. See `delayTerms` variable for rounds along the code.

contracts/Court.sol
```diff
@@ -141,7 +166,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
 
     uint64 internal constant ZERO_TERM_ID = 0; // invalid term that doesn't accept disputes
     uint64 internal constant MODIFIER_ALLOWED_TERM_TRANSITIONS = 1;
-    bytes4 private constant ARBITRABLE_INTERFACE_ID = 0xabababab; // TODO: interface id
+    //bytes4 private constant ARBITRABLE_INTERFACE_ID = 0xabababab; // TODO: interface id
```
izqui: Did this get commented out because of code size limits?

contracts/Court.sol
```diff
@@ -776,49 +899,116 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+        )
     {
+        CourtConfig storage config = courtConfigs[terms[_currentRound.draftTermId].courtConfigId];
+        require(_roundId < config.maxRegularAppealRounds, ERROR_INVALID_ADJUDICATION_ROUND);
+
+        appealDraftTermId = _endTermForAdjudicationRound(_currentRound);
+
+        if (_roundId == config.maxRegularAppealRounds - 1) { // final round, roundId starts at 0
+            // number of jurors will be the number of times the minimum stake is hold in the tree, multiplied by a precision factor for division roundings
+            appealJurorNumber = _getFinalAdjudicationRoundJurorNumber();
```
izqui: This number will vary depending on when the function is called as it will always return the total number of active tokens for the current `termId`. This is a bit problematic, as the deposit amount can change between `appeal`, `appealConfirm` and `settleAppealDeposit`.

The termId that we use for calculating the fees for the final round should be pegged to a constant term id (likely the first term in which the round could be appealed?). We would have to also use this term as the checkpoint for balances when voting in the final round.

It does feel a bit weird, but I don't know how else we could prevent having an imbalance between fees and actual jurors voting.

contracts/Court.sol
bingen: Moved to #96

contracts/Court.sol
bingen: Well, more or less. Not really big gain, but it wasn't used either, right?

contracts/Court.sol
bingen: Moved to #96

contracts/Court.sol
bingen: Oh, I see where now. the problem is that I'm afraid this was introduced with merge of `development` branch. It was not repeated on commit 03a4621 (last one before that merge).
See: https://github.com/aragon/aragon-court/blob/03a462158de186266f258dc94688a03d1fa6435f/contracts/Court.sol#L495
Could you double check that merge to see there was nothing else and maybe revert that particular change?

contracts/Court.sol
bingen: Yes, there's actually a comment about it below, in line 963: 
https://github.com/aragon/aragon-court/blob/b71a78c5d7cb58869e120436b5d0b09e7d536917/contracts/Court.sol#L963
I think I would open a separate issue for this (as this is actually independent of this PR).

bingen: Created issue #97

bingen: Created issue #97

contracts/Court.sol
```diff
@@ -773,49 +895,115 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+            uint256 appealDeposit,
+            uint256 appealConfirmDeposit
+        )
     {
+        CourtConfig storage config = courtConfigs[terms[_currentRound.draftTermId].courtConfigId];
+        require(_roundId < config.maxRegularAppealRounds, ERROR_INVALID_ADJUDICATION_ROUND);
+
+        appealDraftTermId = _endTermForAdjudicationRound(_currentRound);
+
+        if (_roundId == config.maxRegularAppealRounds - 1) { // final round, roundId starts at 0
```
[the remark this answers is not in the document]
bingen: I don't think we need this, as we have:
```
        require(_roundId < config.maxRegularAppealRounds, ERROR_INVALID_ADJUDICATION_ROUND);
```
2 lines above.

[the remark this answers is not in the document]
bingen: Oh, I see, the configurable max regular appeal parameter has been already merged. Is that the reason? Then I guess we should get rid of that `require`? Are we sure this would work? It looks weird to me, I'll have a look to the other PR, I don't recall it too much right now.

contracts/Court.sol
bingen: Well, yes, but if there's no appeal, the dispute is over, so it doesn't really matter. This is only internally called by `_getNextAppealDetails`, which in turn is only called using `appealDraftTermId` in `appealConfirm`, so it's not an issue.

contracts/Court.sol
izqui: No, not yet

contracts/Court.sol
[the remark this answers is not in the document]
izqui: We decided on putting it in the regular configuration for now, and see whether we want to 'cache' some settings at the dispute level later (other settings as the term duration for the adjudication stages are equally weird if they are modified during the dispute 😉)

contracts/Court.sol
izqui: Sounds good. Let's solve it in the other issue

PR #92, opened 2019-08-06
 HexSumTree: Checkpoint root depth

izqui:


contracts/lib/HexSumTree.sol
```diff
@@ -9,7 +9,7 @@ library HexSumTree {
 
     struct Tree {
         uint256 nextKey;
-        uint256 rootDepth;
+        Checkpointing.History rootDepth;
```
facuspagnuolo: should we rename this to simply `root` now?

facuspagnuolo: If so, I'd rename the `getRootDepthX` to simply `getRootX` following what we do for `nodes` and `getItemX` getters

contracts/lib/HexSumTree.sol
```diff
@@ -149,6 +156,14 @@ library HexSumTree {
         return self.nodes[INSERTION_DEPTH][key].get(time);
     }
 
+    function getRootDepth(Tree storage self) internal view returns (uint256) {
+        return self.rootDepth.getLast();
+    }
+
+    function getRootDepthAt(Tree storage self, uint64 time, bool past) internal view returns (uint256) {
+        return past ? self.rootDepth.get(time) : self.rootDepth.getLastPresent(time);
```
facuspagnuolo: Maybe we can split this to avoid an unnecessary if for `totalSumPresent` and `totalSumPast` if you already know which function you need to use for each case

izqui: Maybe I am not following properly, but there are functions that pass a boolean to this function that is gotten as an argument.

https://github.com/aragon/aragon-court/pull/92/files#diff-87cd8c9cdbe46015e73866e9f125bc9dR80-R81

Would you rather have the if in these functions?

contracts/lib/HexSumTree.sol
izqui: I think I'd rather keep the `rootDepth` name as it isn't a node but the depth of the tree.

contracts/lib/HexSumTree.sol
facuspagnuolo: I messed up with the `past` param, I thought this was called only from `totalSumPresent` and `totalSumPast`

PR #96, opened 2019-08-07
Voting: Improve implementation and document

facuspagnuolo:
This PRs proposes a few improvements regarding the voting app:

- Drop `Ruling` enum, in favor of modeling missing, leaked, and refused votes. Now we have a more robust outcomes/rulings validation. 
- Fix the fact that leaked votes were considered as refused, which may lead to rewarding some fees unintentionally in some edge cases.
- Cache jurors' weight computation for the final appeal round. If we don't do this, we will be basically collecting tokens from jurors more than once or cannot be actually slashed for leaked votes.
- Trust always the voting app about the winning outcome
- Jurors that commit a refused vote in the final appeal round are getting slashed
- Appeals must be in favor of valid outcomes 
- Improve old `canCommit` and `canReveal` functions
- Clarify how jurors' weight is computed for final rounds
- Clarify and improve `CRVoting` code
- New terminology
  - Use `Voting` instead of `Vote` to refer to voting instances
  - Use `Vote` instead of `CastVote` to refer to a voter's decision on a `Voting`
  - Use `outcome` instead of `ruling` within the voting app itself
- Inline documentation related to voting functionality


bingen:
> We should revisit this concept if we want to introduce the idea of refusing to vote.

What do you mean? This idea is already introduced.

> Fix the fact that leaked votes were considered as refused, which may lead to rewarding some feeds unintentionally in some edge cases.

What do you mean with "rewarding some feeds unintentionally"? 

> Trust always the voting app about the winning outcome

What does this mean?

> Jurors that commit a refused vote in the final appeal round are getting slashed

I don't think this is right. Jurors could refuse for instance because they think there's not enough evidence or something like that. cc @lkngtn @izqui 

> Appeals must be in favor of valid outcomes

Wasn't this the case before? (Then it was a bug)

> Use Voting instead of Vote to refer to voting instances

We don't do that in our regular Voting app. Actually, I remember long time ago proposing this same change to that app and several native English speakers telling me I was wrong, hahaha.

> Use Vote instead of CastVote to refer to a voter's decision on a Voting

If we revert previous change, I would revert this one too, to avoid confusion (that's why we use `CastVote`). We use `CastVote` in our regular voting app too.

> Use outcome instead of ruling within the voting app itself

I don't mind this one that much, but `CRVoting` app is not meant to be general purpose or reusable, but tightly coupled to `Court`, so I don't think it's a problem to use terminology from Court. Actually the opposite, it makes it easier to follow the code and reason about in my opinion.

contracts/CRVoting.sol
```diff
@@ -1,191 +1,294 @@
...
+        bool leaked;            // whether the outcome casted was leaked or not
     }
 
-    enum Ruling {
-        Missing,
-        Refused
-        // ruling options are dispute specific
+    struct Voting {
+        uint8 winningOutcome;                       // outcome winner of a voting
+        uint8 possibleOutcomes;                     // number of possible outcomes that can be voted
```
bingen: I don't think we need to allow this to be variable. This contract is tightly coupled to Court, and that one wouldn't work with more than 2 outcomes.
I know this was already here before (with different name), but I think we should remove it and make it a constant like `MIN_OUTCOMES_ALTERNATIVES`.

contracts/CRVoting.sol
```diff
@@ -1,191 +1,294 @@
...
+    * @dev Internal function to tell whether a certain outcome is valid for a given voting or not. Note that
+    *      the outcome zero is not considered valid, it is used to denote absence. This function assumes
+    *      the given voting exists.
+    * @param _votingId ID of the voting to check the outcome of
+    * @param _outcome Outcome to check if valid or not
+    * @return True if the given outcome is valid for the requested voting, false otherwise.
+    */
+    function _isValidOutcome(uint256 _votingId, uint8 _outcome) internal view returns (bool) {
+        Voting storage voting = votingRecords[_votingId];
+        return _outcome > uint256(0) && _outcome <= voting.possibleOutcomes;
```
bingen: What's this `0` here? It would be nice to have it as a constant instead of hardcoded here.

contracts/Court.sol
```diff
@@ -664,45 +643,52 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+        require(round.settledPenalties, ERROR_ROUND_NOT_SETTLED);
+        require(jurorState.weight > uint256(0), ERROR_INVALID_JUROR);
         jurorState.rewarded = true;
 
-        uint256 voteId = _getVoteId(_disputeId, _roundId);
-        uint256 coherentJurors = round.coherentJurors;
-        uint8 jurorRuling = voting.getCastVote(voteId, _juror);
+        // Check the given juror is considered a winner based on his vote.
+        uint256 votingId = _getVotingId(_disputeId, _roundId);
+        require(voting.isWinningVoter(votingId, _juror), ERROR_JUROR_NOT_COHERENT);
```
izqui: 🚨 The round slashing/rewarding should take into account the final outcome of the dispute, not the round in particular, so if a ruling is appealed and the final ruling is flipped, those jurors need to be slashed.

contracts/CRVoting.sol
```diff
@@ -1,191 +1,343 @@
...
+    /**
+    * @dev Internal function to tell whether a certain outcome is valid for a given voting or not. Note that
+    *      the missing and leaked outcomes are not considered valid. The only outcomes considered valid are
+    *      refused or any of the possible outcomes of the given voting. This function assumes the given voting exists.
+    * @param _voting Pointer to the voting to check the outcome of
+    * @param _outcome Outcome to check if valid or not
+    * @return True if the given outcome is valid for the requested voting, false otherwise.
+    */
+    function _isValidOutcome(Voting storage _voting, uint8 _outcome) internal view returns (bool) {
+        return _outcome > OUTCOME_REFUSED && _outcome <= _voting.maxAllowedOutcome;
```
izqui: I think we should consider making 'refused' a valid outcome, otherwise a juror voting for 'refused' will never be penalized.

contracts/Court.sol
```diff
@@ -577,42 +555,47 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
 
+        // Burn tokens and refund fees only if we finished settling all the jurors that voted in this round
         if (round.settledPenalties) {
-            // No juror was coherent in the round
+            uint256 collectedTokens = round.collectedTokens;
+            emit RoundSlashingSettled(_disputeId, _roundId, collectedTokens);
+
+            // Avoid calling the jurors registry if there were non tokens collected to burn
+            if (collectedTokens > 0) {
+                jurorsRegistry.burnTokens(collectedTokens);
```
izqui: We only burn tokens if no jurors in the round voted for the final ruling of the dispute

contracts/Court.sol
```diff
@@ -664,45 +643,52 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
             // TODO: stack too deep
             penalties[i] = _pct4(jurorsRegistry.minJurorsActiveBalance(), _penaltyPct) * _round.jurorSlotStates[juror].weight;
         }
-        uint8[] memory castVotes = voting.getCastVotes(_voteId, jurors);
-        // we assume:
-        //require(castVotes.length == batchSettledJurors);
-        collectedTokens = jurorsRegistry.slashOrUnlock(termId, jurors, penalties, castVotes, _winningRuling);
+        bool[] memory losingJurors = voting.getLosingVoters(_votingId, jurors);
```
izqui: Here we'd also need to input voting what the final ruling was, as a juror could have been in the losing side in their round, but in the winning side in the final ruling (as a result of an appeal)

contracts/Court.sol
izqui: I think right now we are both burning tokens and distributing their part to winning jurors (spending 2x the amount of collected tokens)

contracts/Court.sol
```diff
@@ -784,96 +770,110 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+    function _getJurorWeightForFinalRound(uint256 _disputeId, uint256 _roundId, address _juror) internal returns (uint256) {
+        AdjudicationRound storage round = disputes[_disputeId].rounds[_roundId];
+        JurorState storage jurorState = round.jurorSlotStates[_juror];
 
-        uint256 minJurorsActiveBalance = jurorsRegistry.minJurorsActiveBalance();
-        // weight is the number of times the minimum stake the juror has, multiplied by a precision factor for division roundings
-        uint256 stake = jurorsRegistry.activeBalanceOfAt(_voter, disputes[_disputeId].rounds[_roundId].draftTermId);
-        if (stake < minJurorsActiveBalance) {
-            return 0;
+        // If the juror weight for the las round was already computed, return that value
```
izqui:
```suggestion
        // If the juror weight for the last round was already computed, return that value
```

contracts/CRVoting.sol
facuspagnuolo: The problem is that you will force the court to handle binary disputes forever, I'd rather see how much gas we sacrifice with this and decide then

contracts/CRVoting.sol
facuspagnuolo: I think this is an old comment, it was already fixed

contracts/CRVoting.sol
facuspagnuolo: I thought the idea of refusing to vote was actually to avoid participating in a dispute that a juror may not have full information to act, should they get slashed for that?

contracts/Court.sol
facuspagnuolo: > We only burn tokens if no jurors in the round voted for the final ruling of the dispute

I don't think this is what we're doing. I created a new issue to fix the burn tokens accounting (see https://github.com/aragon/aragon-court/issues/99)

contracts/CRVoting.sol
bingen: About the gas, I guess it's an SSTORE for each vote plus some SLOADS, so not negligible, but I don't think it matters, as handling non-binary disputes would increase complexity and require changes in the code of Court, which is not an upgradeable contract.

contracts/CRVoting.sol
bingen: Sorry, I guess I didn't refresh this.

contracts/CRVoting.sol
bingen: I think they should only be slashed if it's not the what the majority of the jurors voted. But if majority of jurors decide to refuse, it should be ok.

facuspagnuolo: Cool, let's do that

contracts/Court.sol
facuspagnuolo: I closed the issue, I mess things up when I moved this statement outside of the `if (round.coherentJurors == 0)`

contracts/Court.sol
facuspagnuolo: Good catch, I completely missed that. Fixed in https://github.com/aragon/aragon-court/pull/96/commits/a90249763aea79f6d853b52ad66f4dd6ed9f6480

contracts/CRVoting.sol
facuspagnuolo: Fixed in https://github.com/aragon/aragon-court/pull/96/commits/aaac2d371e72b1c70117bbf088d66a5f50c5e0eb

contracts/Court.sol
facuspagnuolo: Good catch, I completely missed that. Fixed in https://github.com/aragon/aragon-court/pull/96/commits/a90249763aea79f6d853b52ad66f4dd6ed9f6480

contracts/Court.sol
facuspagnuolo: https://github.com/aragon/aragon-court/pull/96/commits/fe968081d1b0b5fb9920114554924100c02ed31d

contracts/Court.sol
```diff
@@ -657,12 +629,14 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
             penalties[i] = _pct4(jurorsRegistry.minJurorsActiveBalance(), _penaltyPct) * _round.jurorSlotStates[juror].weight;
         }
-        uint8[] memory castVotes = voting.getCastVotes(_voteId, jurors);
-        // we assume:
-        //require(castVotes.length == batchSettledJurors);
-        collectedTokens = jurorsRegistry.slashOrUnlock(termId, jurors, penalties, castVotes, _winningRuling);
 
+        // Check which of the batch of jurors voted in favor of the final ruling of the dispute in this round
+        // we assume `jurorsInFavor` length is equal to `batchSettledJurors`
+        bool[] memory jurorsInFavor = voting.getVotersInFavorOf(_votingId, _finalRuling, jurors);
```
facuspagnuolo: > Trust always the voting app about the winning outcome

@bingen I meant this for example, we can ask the voting app who voted in favor of the final ruling instead of filtering that ourselves in the registry. 

You can also see that we are trusting the voting app [here](https://github.com/aragon/aragon-court/pull/96/files#diff-4bb317b489994cf6f4c4b7ed11c18590R429) and [here](https://github.com/aragon/aragon-court/pull/96/files#diff-4bb317b489994cf6f4c4b7ed11c18590R453)  to ask if a given ruling is valid or not.

contracts/CRVoting.sol
```diff
@@ -1,191 +1,354 @@
...
     /**
-     * @notice Leak vote for `_voter` in vote #`_voteId`
-     */
-    function leakVote(uint256 _voteId, address _voter, uint8 _leakedRuling, bytes32 _salt) external {
-        Vote storage vote = votes[_voteId];
-        require(vote.possibleRulings > 0, ERROR_OUT_OF_BOUNDS);
-        uint256 weight = owner.canCommit(_voteId, _voter);
-        require(weight > 0, ERROR_NOT_ALLOWED_BY_OWNER);
-
-        uint8 ruling = uint8(Ruling.Refused);
```
facuspagnuolo: > Fix the fact that leaked votes were considered as refused, which may lead to rewarding some fees unintentionally in some edge cases.

@bingen as you can see, leaked votes were considered refused. If the final outcome of a round was actually refused, leaked voters were getting rewarded.

contracts/CRVoting.sol
```diff
@@ -1,191 +1,354 @@
...
+    /**
+    * @dev Internal function to tell whether a certain outcome is valid for a given voting or not. Note that
+    *      the missing and leaked outcomes are not considered valid. The only outcomes considered valid are
+    *      refused or any of the possible outcomes of the given voting. This function assumes the given voting exists.
+    * @param _voting Pointer to the voting to check the outcome of
+    * @param _outcome Outcome to check if valid or not
+    * @return True if the given outcome is valid for the requested voting, false otherwise.
+    */
+    function _isValidOutcome(Voting storage _voting, uint8 _outcome) internal view returns (bool) {
+        return _outcome >= OUTCOME_REFUSED && _outcome <= _voting.maxAllowedOutcome;
```
facuspagnuolo: > Jurors that commit a refused vote in the final appeal round are getting slashed

@bingen we are considering refused a valid outcome, which means that if a juror refused to vote but refused is not the winning outcome, he will get slashed

contracts/Court.sol
```diff
@@ -424,65 +417,45 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
-        // check correct ruling
-        require(
-            _forRuling > uint8(Ruling.Refused) &&
-            _forRuling <= uint8(Ruling.Refused) + MAX_RULING_OPTIONS &&
-            _forRuling != currentRuling,
-            ERROR_INVALID_RULING
-        );
+        uint256 votingId = _getVotingId(_disputeId, _roundId);
+        uint8 roundWinningRuling = voting.getWinningOutcome(votingId);
+        require(roundWinningRuling != _ruling && voting.isValidOutcome(votingId, _ruling), ERROR_INVALID_APPEAL_RULING);
```
facuspagnuolo: > Appeals must be in favor of valid outcomes

@bingen we are now trusting the voting app to check whether an outcome is valid or not. Discussing with @izqui we arrived to the idea that appeal in favor of the "refused" outcome is something valid. A juror could insist that there is not enough evidence to decide on a dispute for example.

contracts/CRVoting.sol
```diff
@@ -1,191 +1,354 @@
...
+    // Outcome nr. 0 is used to denote a missing vote (default)
+    uint8 internal constant OUTCOME_MISSING = uint8(0);
+    // Outcome nr. 1 is used to denote a leaked vote
+    uint8 internal constant OUTCOME_LEAKED = uint8(1);
+    // Outcome nr. 2 is used to denote a refused vote
+    uint8 internal constant OUTCOME_REFUSED = uint8(2);
+    // Besides the options listed above, every voting must provide at least 2 outcomes
+    uint8 internal constant MIN_POSSIBLE_OUTCOMES = uint8(2);
+    // Max number of outcomes excluding the default ones
+    uint8 internal constant MAX_POSSIBLE_OUTCOMES = uint8(-1) - OUTCOME_REFUSED;
```
facuspagnuolo: > F: We should revisit this concept if we want to introduce the idea of refusing to vote.
> B: What do you mean? This idea is already introduced.

@bingen Yes, it was introduced, but not fully modeled, I think you can see we have now proper outcomes to denote missing, leaked, refused or any of the custom ones in the voting app

facuspagnuolo:
@bingen I hope I've answered all your questions, I drop some answers in separate threads cause GH doesn't allow to start a thread in general comments :/

The only thing I didn't point out in the code is terminology, I don't fully get if you are saying that you proposed it but native speakers warned you that it was wrong, or if they warned that what we have now is wrong and you proposed to change it.

contracts/CRVoting.sol
bingen: Good catch!

contracts/CRVoting.sol
bingen: Ok, although I think refused already was a valid outcome (as I said, there even was a test for it in the collateral appeal PR)

contracts/Court.sol
bingen: As I said in the previous one, refused was already valid.

contracts/CRVoting.sol
bingen: Two comments here:

- As I said before, I think we should simplify everything to allow only for binary disputes.
- Here you are also relying on the "position" of `OUTCOME_REFUSED`. See: https://github.com/aragon/aragon-court/pull/75#discussion_r311261880

bingen:
> I don't fully get if you are saying that you proposed it but native speakers warned you that it was wrong

Yes, I proposed the same change as you (Vote -> Voting) and they warned me I was wrong. :wink: Fun fact, if you dig deep down to the prehistory of Aragon, you'll see votes were called votings at some point (too lazy to search for it now).

contracts/CRVoting.sol
facuspagnuolo: The problem was that you could refuse to vote upon an appeal, but not appeal in favor of the refused outcome.

contracts/Court.sol
facuspagnuolo: Maybe this snippet is not showing the full picture, please see https://github.com/aragon/aragon-court/pull/96/files/b97df5ce134620d0692f56ee185b5a003e8ecd83#diff-4bb317b489994cf6f4c4b7ed11c18590L437-L442

contracts/CRVoting.sol
facuspagnuolo: Sure, but internally, note that the Court contract doesn't know anymore how the CRVoting contract deals with refused votes

contracts/CRVoting.sol
```diff
@@ -1,191 +1,354 @@
...
+    *      outcome, it means that no one voted in favor of any of the possible outcomes.
+    * @param _votingId ID of the voting to query if a voter voted in favor of a certain outcome
+    * @param _outcome Outcome to query if the given voter voted in favor of
+    * @param _voter Address of the voter to query if voted in favor of the given outcome
+    * @return True if the given voter voted in favor of the given outcome, false otherwise
+    */
+    function hasVotedInFavorOf(uint256 _votingId, uint8 _outcome, address _voter) external votingExists(_votingId) view returns (bool) {
+        Voting storage voting = votingRecords[_votingId];
+        uint8 winningOutcome = voting.winningOutcome;
+        return winningOutcome != OUTCOME_MISSING && voting.votes[_voter].outcome == _outcome;
```
izqui: We should also return false if `_outcome == OUTCOME_MISSING` otherwise we could be rewarding jurors that didn't show up to vote in round 1 (but the outcome of the round wasn't missing) if the winning ruling for round 2 is missing.

In any case, allowing the final ruling of a dispute to be missing is quite weird (and can lead to edge cases like this one). Previously if the winning ruling of a round was missing, we returned `Refused` (since the tally for all the options was a tie to 0, and Refused had the lowest index). I think returning `Refused` would be more consistent although we may have to change some things in this contract.

contracts/Court.sol
```diff
@@ -571,72 +540,75 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
 
-        if (dispute.winningRuling > 0) {
-            return dispute.winningRuling; // winning ruling was already set
+        // Check if there was a final ruling already cached
+        if (dispute.finalRuling > 0) {
+            return dispute.finalRuling;
         }
 
-        // ensure the last round adjudication period already ended
+        // Ensure the last adjudication round has ended. Not there will always be at least one round.
```
izqui:
```suggestion
        // Ensure the last adjudication round has ended. Note that there will always be at least one round.
```

contracts/Court.sol
```diff
@@ -777,96 +751,110 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+    function _getJurorWeightForFinalRound(uint256 _disputeId, uint256 _roundId, address _juror) internal returns (uint256) {
+        AdjudicationRound storage round = disputes[_disputeId].rounds[_roundId];
+        JurorState storage jurorState = round.jurorSlotStates[_juror];
 
-        uint256 minJurorsActiveBalance = jurorsRegistry.minJurorsActiveBalance();
-        // weight is the number of times the minimum stake the juror has, multiplied by a precision factor for division roundings
-        uint256 stake = jurorsRegistry.activeBalanceOfAt(_voter, disputes[_disputeId].rounds[_roundId].draftTermId);
-        if (stake < minJurorsActiveBalance) {
-            return 0;
+        // If the juror weight for the las round was already computed, return that value
```
izqui:
```suggestion
        // If the juror weight for the last round was already computed, return that value
```

contracts/Court.sol
```diff
@@ -777,96 +751,110 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+        CourtConfig storage config = courtConfigs[terms[draftTermId].courtConfigId];
+        uint256 weightedPenalty = _pct4(activeBalance, config.penaltyPct);
+        if (!jurorsRegistry.collectTokens(_juror, weightedPenalty, termId)) {
+            // If it was not possible to collect the amount to be locked, return 0 to prevent juror from voting
+            return uint256(0);
+        }
 
-        return _getJurorWeight(_disputeId, _roundId, _voter);
+        // If it was possible to collect the amount of active tokens to be locked, update the final round state
+        uint256 weight = FINAL_ROUND_WEIGHT_PRECISION.mul(activeBalance / minJurorsActiveBalance);
```
izqui:
```suggestion
        uint256 weight = FINAL_ROUND_WEIGHT_PRECISION.mul(activeBalance) / minJurorsActiveBalance;
```

We multiply by `FINAL_ROUND_WEIGHT_PRECISION` to avoid losing precision in `activeBalance / minJurorsActiveBalance`

contracts/Court.sol
```diff
@@ -777,96 +751,110 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
+        jurorState.weight = weight;
+        round.collectedTokens = round.collectedTokens.add(weightedPenalty);
+        return weight;
     }
 
     function getJurorWeight(uint256 _disputeId, uint256 _roundId, address _juror) external view returns (uint256) {
-        return _getJurorWeight(_disputeId, _roundId, _juror);
+        return _getJurorWeightForRegularRound(_disputeId, _roundId, _juror);
```
izqui: Why not `_getJurorWeight`?

contracts/CRVoting.sol
bingen: Oh, now I get your original comment. Yes, this was definitely something pending, as noted here: https://github.com/aragon/aragon-court/blob/development/contracts/Court.sol#L92

contracts/Court.sol
facuspagnuolo: Cause we have the other one called `_getJurorWeightForFinalRound`

contracts/CRVoting.sol
facuspagnuolo: yup :)

contracts/CRVoting.sol
facuspagnuolo: This is assuming there won't be more than 2^64 weight in favor of an outcome, which could be possible in an edge case for the last appeal round, am I right?

contracts/Court.sol
```diff
@@ -571,72 +540,75 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
...
 
-        uint8 winningRuling = _ensureFinalRuling(_disputeId);
-        uint256 voteId = _getVoteId(_disputeId, _roundId);
-        // let's fetch them only the first time
+        // Set the number of jurors that voted in favor of the final ruling if we haven't started settling yet
+        uint8 finalRuling = _ensureFinalRuling(_disputeId);
+        uint256 votingId = _getVotingId(_disputeId, _roundId);
         if (round.settledJurors == 0) {
-            round.coherentJurors = uint64(voting.getRulingVotes(voteId, winningRuling));
+            // TODO: review casting, could this overflow?
```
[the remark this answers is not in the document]
facuspagnuolo: This is assuming there won't be more than 2^64 weight in favor of an outcome, which could be possible in an edge case for the last appeal round, am I right?

Regarding the `uint256` for the jurors, I decided to roll it back and create a new issue to add a validation in the registry

contracts/Court.sol
```diff
@@ -64,7 +64,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
     }
 
     struct JurorState {
-        uint64 weight;
+        uint256 weight;
```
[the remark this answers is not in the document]
facuspagnuolo: I decided to roll it back and create a new issue to add a validation in the registry (see https://github.com/aragon/aragon-court/issues/102)

contracts/Court.sol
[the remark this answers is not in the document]
izqui: Yes, which means that the difference in magnitude between the total active balance and the minimum active balance can be up to 1.84e16 (taking into account the precision multiplier).

Unless my math is off I think we should be fine (as long as we are not doing math in uint64 which can make an intermediate value overflow).

contracts/Court.sol
```diff
@@ -898,18 +886,6 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner {
         }
     }
 
-    function _getRoundWinningRuling(uint256 _disputeId, uint256 _roundId) internal view returns (uint8 winningRuling) {
```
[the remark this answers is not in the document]
facuspagnuolo: True

contracts/Court.sol
izqui: Yes, but there is no external function (other than `canCommit`) to get the weight a juror has in the final round as this function only returns the weight for regular rounds. As `_getJurorWeightForFinalRound` now modifies state I assume that's the reason not to do it?

Actually after the juror has committed their vote in the final round it will return their cached weight (before that it will return 0)

facuspagnuolo: Exactly. I'll improve it to at least compute the weight a juror will have for the final round before collecting tokens

facuspagnuolo:
@bingen @izqui for reference, I reverted the terminology changes [here](https://github.com/aragon/aragon-court/pull/98/commits/0df7ae0987ac90895f83a5bf540cf3f9d0dfab61)

contracts/CRVoting.sol
```diff
@@ -1,191 +1,355 @@
...
-    function getRulingVotes(uint256 _voteId, uint8 _ruling) external view returns (uint256) {
-        Vote storage vote = votes[_voteId];
-        require(vote.possibleRulings > 0, ERROR_OUT_OF_BOUNDS);
+    /**
+    * @dev Internal function to fetch and ensure the weight of voter willing to commit a vote for a certain voting
+    * @param _votingId ID of the voting to check the voter's weight of
+    * @param _voter Address of the voter willing to commit a vote
+    * @return Weight of the voter willing to commit a vote
+    */
+    function _ensureVoterWeightToCommit(uint256 _votingId, address _voter) internal returns (uint256) {
```
bingen: About all these `_ensure...` functions, see [this comment](https://github.com/aragon/aragon-court/pull/103#discussion_r321805531)

contracts/CRVoting.sol
```diff
@@ -1,191 +1,355 @@
...
+        return winningOutcome == OUTCOME_MISSING ? OUTCOME_REFUSED : winningOutcome;
+    }
 
-        winningRuling = vote.winningRuling;
+    /**
+    * @dev Get the tally of the winning outcome for a certain voting
+    * @param _votingId ID of the voting querying the tally of
+    * @return Tally of the winning outcome being queried for the given voting
+    */
+    function getWinningOutcomeTally(uint256 _votingId) external votingExists(_votingId) view returns (uint256) {
```
bingen: Where is this being used? I can't find it. If we don't need it, at least I would remove it from the interface.

PR #104, opened 2019-08-23
Tree: Polish contract and write full-coverage tests

### LINKED ISSUE #94: Write full-coverage unit tests for HexSumTree
Please follow the convention used for the `JurorsRegistry` tests, some of them are:
- Avoid mocking unnecessarily, only when there is no other option
- Use `test-helpers` module provided by `@aragon`
- Implement reusable helpers stuff instead of custom non-exported functions
- Use `let` scoped variables instead of properties to define test variables

facuspagnuolo:
Fixes #94 

This PR introduces some core changes in the `HexSumTree` lib functionality:
- Drop lots of unused functionality (https://github.com/aragon/aragon-court/pull/104/commits/4dfd51a6978ff3717b256f76477530e37a34395b)
- Work with inclusive bounds for batched drafting and handle roundings ([R63-R74](https://github.com/aragon/aragon-court/pull/104/commits/3caed2c5faec71b23b0cfa0761e60d3aaebf964b#diff-cf115e6db6bfff22baed3838552cc035R63-R74))
- Precompute randomness seed with draft constants ([R39](https://github.com/aragon/aragon-court/pull/104/commits/3caed2c5faec71b23b0cfa0761e60d3aaebf964b#diff-cf115e6db6bfff22baed3838552cc035R39))
- Include equal balance scenarios in the tree search ([R410](https://github.com/aragon/aragon-court/pull/104/commits/3caed2c5faec71b23b0cfa0761e60d3aaebf964b#diff-87cd8c9cdbe46015e73866e9f125bc9dR410))


contracts/lib/HexSumTree.sol
```diff
@@ -27,100 +28,113 @@ import "./Checkpointing.sol";
...
-        self.nextKey++;
+        uint256 key = self.nextKey++;
 
+        // If the new value is not zero, first set the value of the new leaf node, then add a new level at the top of
+        // the tree if necessary, and finally update sums cached in all the non-leaf nodes.
         if (_value > 0) {
-            _set(self, key, _time, _value);
+            // TODO: tree height is never checked to be under 64, at least add a comment telling why
+            _add(self, LEAVES_LEVEL, key, _time, _value);
+            _addLevelIfNecessary(self, key, _time);
```
izqui: We should add the new level regardless of whether the insertion value was greater than 0, otherwise a node could be inserted with an initial value of 0, then set to another value which will not be included in the total sum.

contracts/lib/HexSumTree.sol
```diff
@@ -27,100 +28,113 @@ import "./Checkpointing.sol";
...
 
+    string private constant ERROR_UPDATE_OVERFLOW = "SUM_TREE_UPDATE_OVERFLOW";
+    string private constant ERROR_KEY_DOES_NOT_EXIST = "SUM_TREE_KEY_DOES_NOT_EXIST";
+
+    // Constants used to perform tree computations
+    uint256 private constant CHILDREN = 16;
+    uint256 private constant BITS_IN_NIBBLE = 4;
+
+    // Leaves are at height or level zero, root height will be increasing as new levels are inserted in the tree
+    uint256 private constant LEAVES_LEVEL = 0;
```
izqui: Super nitpicky: all upper nodes have leaves as well. How about calling it `VALUES_LEVEL`/`ITEMS_LEVEL`?

contracts/lib/HexSumTree.sol
```diff
@@ -27,100 +28,113 @@ import "./Checkpointing.sol";
...
-        require(key <= self.nextKey, ERROR_NEW_KEY_NOT_ADJACENT); // TODO: change to strictly <
-        _set(self, key, time, value);
+    function set(Tree storage self, uint256 _key, uint64 _time, uint256 _value) internal {
+        require(_key < self.nextKey, ERROR_KEY_DOES_NOT_EXIST);
+
+        // Set the new value for the requested leaf node
+        uint256 lastValue = getItem(self, _key);
+        _add(self, LEAVES_LEVEL, _key, _time, _value);
+
+        // Update sums cached in the non-leaf nodes. Note that overflows is being checked at the end of the whole update.
```
izqui:
```suggestion
        // Update sums cached in the non-leaf nodes. Note that overflows are being checked at the end of the whole update.
```

contracts/lib/HexSumTree.sol
```diff
@@ -210,57 +268,70 @@ library HexSumTree { // TODO: rename to CheckpointedHexSumTree?
...
+    * @param _delta Numeric delta to update the ancestors' values of the given leaf node
+    * @param _positive Boolean to tell whether the given delta should be added to or subtracted from ancestors' values
+    */
+    function _updateSums(Tree storage self, uint256 _key, uint64 _time, uint256 _delta, bool _positive) private {
         uint256 mask = uint256(-1);
-        uint256 ancestorKey = key;
-        for (uint256 i = 1; i <= currentRootDepth; i++) { // TODO: rename i to level and currentRootDepth to height
+        uint256 ancestorKey = _key;
+        uint256 currentHeight = getHeight(self);
+        for (uint256 level = 1; level <= currentHeight; level++) {
```
izqui:
```suggestion
        for (uint256 level = LEAVES_LEVEL + 1; level <= currentHeight; level++) {
```

contracts/lib/HexSumTree.sol
```diff
@@ -422,52 +390,43 @@ library HexSumTree { // TODO: rename to CheckpointedHexSumTree?
...
-            }
+    /**
+    * @dev Private function to tell how many values of a list can be found in a subtree
+    * @param _values List of values being searched in ascending order
+    * @param _foundValues Number of values that were already found and should be ignore
+    * @param _subtreeTotal Total sum of the given subtree to check the numbers that are included in it
+    * @return Number of values in the list that are included in the given subtree
+    */
+    function _getValuesIncludedInSubtree(uint256[] memory _values, uint256 _foundValues, uint256 _subtreeTotal) private pure returns (uint256) {
+        uint256 i = _foundValues;
```
izqui: We should exit early if the subtree doesn't contain any value:

```
if (_subtreeTotal == 0) {
    return 0;
}
```

This will also prevent selecting the key 0 (inserted by the Registry on initialization and kept to 0) in the very unlikely case that one of the values being sought is 0.

contracts/lib/JurorsTreeSortition.sol
```diff
@@ -0,0 +1,153 @@
...
+        uint256 _selectedJurors,
+        uint256 _batchRequestedJurors,
+        uint256 _roundRequestedJurors,
+        uint256 _sortitionIteration
+    )
+        internal
+        view
+        returns (uint256[] jurorsIds, uint256[] jurorsBalances)
+    {
+        (uint256 low, uint256 high) = getActiveBalancesBatchBounds(tree, _termId, _selectedJurors, _batchRequestedJurors, _roundRequestedJurors);
```
izqui: Nit: hesitant about using the word `balances` here as we are referring to points in the tree, specially since we are returning `jurorsBalances` that are the actual balances of the jurors selected

contracts/lib/JurorsTreeSortition.sol
```diff
@@ -0,0 +1,153 @@
...
+        uint256 _lowActiveBalanceBatchBound,
+        uint256 _highActiveBalanceBatchBound
+    )
+        internal
+        pure
+        returns (uint256[])
+    {
+        // Calculate the interval to be used to search the balances in the tree. Since we are using a module function
+        // to compute the random balances to be searched, we add one to the difference to make sure the last number
+        // of the range is also included. For example, to compute a range [0,10] we need to compute using module 11.
```
izqui:
```suggestion
        // of the range is also included. For example, to compute a range [0,10] we need to compute using modulo 11.
```

contracts/lib/JurorsTreeSortition.sol
```diff
@@ -0,0 +1,153 @@
...
+        uint256 _selectedJurors,
+        uint256 _batchRequestedJurors,
+        uint256 _roundRequestedJurors
+    )
+        internal
+        view
+        returns (uint256 low, uint256 high)
+    {
+        // Note that the low bound will be always equal to the previous high bound incremented by one, or zero for the
+        // first iteration. Thus, we can make sure we are not excluding any juror from the tree.
```
izqui: If we exclude the point 0 we are not excluding any juror from the tree, actually right now the first juror has 1 more chance of being selected. If we had a tree with 2 jurors, the first one with 1 token and the other one with 2, the interval would be `[0,3]` and the first juror would be selected if the random number is either 0 or 1. IMO we should just skip 0, making the interval `(0, totalActiveBalance]`.

contracts/lib/JurorsTreeSortition.sol
```diff
@@ -0,0 +1,153 @@
...
+    * @param _disputeId Identification number of the dispute to draft jurors for
+    * @param _termId Current term when the draft is being computed
+    * @param _selectedJurors Number of jurors already selected for the draft
+    * @param _batchRequestedJurors Number of jurors to be selected in the given batch of the draft
+    * @param _roundRequestedJurors Total number of jurors requested to be drafted
+    * @param _sortitionIteration Number of sortitions already performed for the given draft
+    * @return jurorsIds List of juror ids obtained based on the requested search
+    * @return jurorsBalances List of active balances for each juror obtained based on the requested search
+    */
+    function multiSortition(
```
izqui:
```suggestion
    function batchedSearch(
```

Nit: Calling this `multiSortition` feels like an abstraction leak, as multi-sortition is how we called the algorithm to perform the sortition for all the values at the same time.

contracts/lib/HexSumTree.sol
```diff
@@ -353,29 +351,29 @@ library HexSumTree {
...
+            uint256 newVisitedTotal = _params.visitedTotal + childNodeValue;
+            uint256 subtreeIncludedValues = _getValuesIncludedInSubtree(_values, _params.foundValues, newVisitedTotal);
 
             // If there are some values included in the subtree of the child node, visit them
             if (subtreeIncludedValues > 0) {
                 // If the child node being analyzed is a leaf, add it to the list of results a number of times equals
                 // to the number of values that were included in it. Otherwise, descend one level
-                if (searchParams.level == 1) {
-                    _copyFoundNode(searchParams.foundValues, subtreeIncludedValues, childNodeKey, resultKeys, childNodeValue, resultValues);
+                if (_params.level == 0) {
```
izqui:
```suggestion
                if (_params.level == LEAVES_LEVEL) {
```

contracts/lib/JurorsTreeSortition.sol
```diff
@@ -0,0 +1,153 @@
+pragma solidity ^0.4.24;
+
+import "./HexSumTree.sol";
+
+
+/**
+* @title JurorsTreeSortition - Library to perform jurors sortition over a `HexSumTree`
+*/
+library JurorsTreeSortition {
```
izqui: 💯

contracts/lib/JurorsTreeSortition.sol
```diff
@@ -0,0 +1,153 @@
...
+        bytes32 _randomnessHash,
+        uint256 _batchRequestedJurors,
+        uint256 _lowActiveBalanceBatchBound,
+        uint256 _highActiveBalanceBatchBound
+    )
+        internal
+        pure
+        returns (uint256[])
+    {
+        // Calculate the interval to be used to search the balances in the tree. Since we are using a module function
```
izqui:
```suggestion
        // Calculate the interval to be used to search the balances in the tree. Since we are using a modulo function
```

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+          await tree.insert(time, 10)
+          assert.equal((await tree.item(1)).toString(), 10, 'value does not match')
+        })
+      })
+
+      context('when adding 40 values', () => {
+        beforeEach('insert 40 values', async () => {
+          // First 16 set of children will be         1^0, 1^1, 1^2, ..., 1^15 at time i+1
+          // Second 16 set of children will be        2^0, 2^1, 2^2, ..., 2^15 at time i+1
+          // Final 8 set of remaining values will be  3^0, 3^1, 3^2, ..., 3^8  at time i+1
```
izqui:
```suggestion
          // Final 8 set of remaining values will be  3^0, 3^1, 3^2, ..., 3^7  at time i+1
```

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+    })
+
+      context('when having 40 values', () => {
+        const insertionTime = 2
+        const setTime = 5
+
+        beforeEach('insert and set 40 values', async () => {
+          // First 16 set of children will be         1^0, 1^1, 1^2, ..., 1^15 at time 2
+          // Second 16 set of children will be        2^0, 2^1, 2^2, ..., 2^15 at time 2
+          // Final 8 set of remaining values will be  3^0, 3^1, 3^2, ..., 3^8  at time 2
```
izqui:
```suggestion
          // Final 8 set of remaining values will be  3^0, 3^1, 3^2, ..., 3^7 at time 2
```

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+
+        context('when the first value is huge', () => {
+          const value = MAX_UINT192
+
+          beforeEach('insert value', async () => {
+            await tree.insert(time, value)
+          })
+
+          it('reverts', async () => {
+            await assertRevert(tree.update(key, time + 1, MAX_UINT256, true), 'SUM_TREE_UPDATE_OVERFLOW')
```
izqui: Wouldn't adding `1` be enough to cause the overflow?

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+        })
+
+        context('when having 40 values', () => {
+          const insertionTime = 2
+          const updateTime = 5
+
+          beforeEach('insert and update 40 values', async () => {
+            // First 16 set of children will be         1^0, 1^1, 1^2, ..., 1^15 at time 2
+            // Second 16 set of children will be        2^0, 2^1, 2^2, ..., 2^15 at time 2
+            // Final 8 set of remaining values will be  3^0, 3^1, 3^2, ..., 3^8  at time 2
```
izqui:
```suggestion
            // Final 8 set of remaining values will be  3^0, 3^1, 3^2, ..., 3^7  at time 2
```

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+    beforeEach('init tree', async () => {
+      await tree.init()
+    })
+
+    context('when searching one value', async () => {
+      const value = 15
+      const searchValues = [value]
+
+      context('when there was no value inserted in the tree', async () => {
+        it('returns zero', async () => {
```
izqui: Shouldn't we just revert in this case as the search was out of bounds? Not concerned about it  when search is used from the `JurorsTreeSortition` lib as we provide proper intervals.

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+
+              assert.equal(keys[0].toString(), 0, 'result key does not match')
+              assert.equal(values[0].toString(), 0, 'result value does not match')
+            })
+          })
+
+          context('when there is a value registered for the given time', async () => {
+            const searchTime = insertTime
+
+            // TODO: this scenario is not passing but it should
```
izqui: Looks like it is passing now

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+
+              assert.equal(keys[0].toString(), 0, 'result key does not match')
+              assert.equal(values[0].toString(), 0, 'result value does not match')
+            })
+          })
+        })
+
+        context('when the total is equal to the searched value', async () => {
+          beforeEach('insert values', async () => {
+            await tree.insert(insertTime, 3)
```
izqui: Can we use some constants for these? A bit hard to follow :)

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+              assert.equal(keys[0].toString(), 5, 'result key does not match')
+              assert.equal(values[0].toString(), 4, 'result value does not match')
+            })
+          })
+        })
+      })
+    })
+
+    context('when searching many values', () => {
+      context('without checkpointing', async () => {
```
izqui: It would be great to parametrize these tests so we can easily create new ones that test things like multi-level sortition (when we have more than 16 insertions). I imagine something like this could be a good input:

```js
const testCases = [
  { insertTime, searchValues: [ 1, 5, 8, ... ], insertValues: [ 2, 1, 4, ...], expectedResults: [[0, 2], [2, 4], [3, 1], ... ]}
]
```

test/tree/hex-sum-tree.js
```diff
@@ -0,0 +1,1116 @@
...
+              assert.equal(values[4].toString(), 6, 'fifth result value does not match')
+
+              assert.equal(keys[5].toString(), 0, 'sixth result key does not match')
+              assert.equal(values[5].toString(), 0, 'sixth result value does not match')
+            })
+          })
+        })
+      })
+
+      context('with checkpointing', async () => {
```
izqui: It would be interesting to test searches at a past time when new checkpoints have been added in the tree.

contracts/lib/HexSumTree.sol
```diff
@@ -152,6 +154,12 @@ library HexSumTree {
      * @return values List of node values found for each requested value in the same order
      */
     function search(Tree storage self, uint256[] _values, uint64 _time) internal view returns (uint256[] keys, uint256[] values) {
+        require(_values.length > uint256(0), ERROR_MISSING_SEARCH_VALUES);
```
izqui: Created an issue to ensure that disputes cannot be created with an initial number of 0 jurors: #104

test/tree/hex-sum-tree.js
```diff
@@ -632,7 +633,7 @@ contract('HexSumTree', () => {
     })
   })
 
-  describe('search', () => {
+  describe.only('search', () => {
```
izqui: tmp remove

test/tree/hex-sum-tree.js
izqui: That would be great. Testing sortition when there are multiple levels in the tree seems useful :)

test/tree/hex-sum-tree.js
izqui: I mean testing to do a search for a single item at a given term in the past when there have been updates in the tree (which shouldn't affect the search thanks to the checkpoints)

PR #108, opened 2019-09-02
Chore: Upgrade Truffle to v5

facuspagnuolo:
**MERGE AFTER #103**

Fixes #82

Issue #23, opened 2019-03-25
Add linter and fix linting issues


izqui:
We use solium in other repos, see aragonOS' config for reference: https://github.com/aragon/aragonOS/blob/dev/.soliumrc.json

bingen:
This is already done [here](https://github.com/aragon/aragon-court/pull/41/commits/37b8d8f541ddf164bf2848187e25b9bf14108441) and [here](https://github.com/aragon/aragon-court/pull/41/commits/d3615ab7acf1534dfa9adad74871cd31b3b5f648).
I will close it when it's merged.

Issue #78, opened 2019-08-06
Court: Explore encapsulating the initialization logic in separate contracts


facuspagnuolo:
Currently, all the initialization logic is triggered by the `Court` constructor.
Explore if we can make things clearer having all these logic in a separate contract (or more than one)

PR #109, opened 2019-09-02
Chore: Upgrade contracts to Solidity 0.5.8

### LINKED ISSUE #80: Migrate contracts codebase to Solidity 0.5
Please explore which patch version should we use having in mind its adoption and stability

facuspagnuolo:
**MERGE AFTER #108**

Fixes #80 and #23

This PR includes:
- Upgrade to solc 0.5.8
- Organize contract files dropping unused mocks and libs
- Add lint task to the CI

I recommend reviewing by commit

⚠️ I decided to upgrade to 0.5.8 since it has been released 4 months ago, and there are only 3 patches on top of it with minor fixes and optimizations. Moreover, I extended some of the contracts from aragonOS to support being compiled with 0.5.x (see [branch](https://github.com/aragon/aragonOS/tree/alpha_upgrade_solidity_5))


contracts/lib/BytesHelpers.sol
```diff
@@ -8,7 +8,15 @@ library BytesHelpers {
 
         if (length < 4) {
```
izqui: As discussed offline, we can remove some of this logic by just returning bytes4(0) if the length is less than 4 bytes.

Also, we could now use `abi.decode` to decode the data (although it doesn't seem to work when the values are not packed :()

contracts/lib/BytesHelpers.sol
```diff
@@ -8,7 +8,15 @@ library BytesHelpers {
...
             uint256 shiftingPositions = 8 - length * 2;
-            result = bytes4(uint256(result) / (16 ** shiftingPositions));
+            result = toBytes4(toUint256(result) / (16 ** shiftingPositions));
         }
     }
+
+    function toBytes4(uint256 i) private pure returns (bytes4 o) {
+        assembly { o := mload(add(i, 0x20)) }
```
izqui: Cannot use `mload` as `i` is already in the stack, `o := i` would do the trick

contracts/standards/arbitration/IArbitrable.sol
```diff
@@ -1,7 +1,7 @@
-pragma solidity ^0.4.15;
+pragma solidity ^0.5.8;
 
 
-interface IArbitrable {
+contract IArbitrable {
```
izqui: What has changed in solidity 0.5 to require this change?

contracts/standards/rng/RNG.sol
```diff
@@ -1,42 +1,44 @@
-/**
-*  @title Random Number Generator Standard
-*  @author Clément Lesaege - <clement@lesaege.com>
-*
-*/
+pragma solidity ^0.5.8;
```
izqui: We could clean up this file now as it is not used anywhere

test/registry/jurors-registry-activatation.js
```diff
@@ -6,12 +6,11 @@ const JurorsRegistry = artifacts.require('JurorsRegistry')
 const ERC20 = artifacts.require('ERC20Mock')
 const JurorsRegistryOwnerMock = artifacts.require('JurorsRegistryOwnerMock')
 
-const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
-
 contract('JurorsRegistry', ([_, juror]) => {
   let registry, registryOwner, ANJ
 
   const MIN_ACTIVE_AMOUNT = bigExp(100, 18)
+  const EMPTY_DATA = '0x0000000000000000000000000000000000000000000000000000000000000000'
```
izqui: Why is this change needed? This is no longer empty data but 32 zero bytes

contracts/standards/rng/RNG.sol
facuspagnuolo: Sure, I already dropped this one

contracts/standards/arbitration/IArbitrable.sol
facuspagnuolo: The thing is that solc 0.5 now requires all interface's methods to be external

contracts/lib/BytesHelpers.sol
facuspagnuolo: Sure, this was a brainfart 😂 
As discussed offline, let's drop this case anyway

test/registry/jurors-registry-activatation.js
facuspagnuolo: Yea, it tests were failing otherwise

facuspagnuolo: I already rolled back this, it was because of the issue with the `BytesHelper` lib. I didn't listen to my tests 😅

facuspagnuolo:
@izqui addressed your comments. I think you will prefer reviewing #108 before merging this one cause it is right in the middle. Let me know

contracts/arbitration/IArbitrable.sol
```diff
@@ -1,7 +1,7 @@
-pragma solidity ^0.4.15;
+pragma solidity ^0.5.8;
 
 
-interface IArbitrable {
```
bingen: Why has this been moved out from `standards` folder?

contracts/standards/arbitration/IArbitrable.sol
bingen: Can't we then make methods external?

contracts/court/Court.sol
```diff
@@ -837,6 +837,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
 
         // Transition the minimum number of terms between the amount requested and the amount actually needed
         uint256 totalFee;
+        CourtConfig storage config = _getConfigSafeAt(termId);
```
bingen: Why do we have 2 calls to `_getConfigSafeAt` now?

contracts/court/Court.sol
```diff
@@ -855,7 +856,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             emit NewTerm(termId, msg.sender);
 
             // Add amount of fees to be paid for the transitioned term
-            CourtConfig storage config = _getConfigSafeAt(currentTerm.courtConfigId);
+            config = _getConfigSafeAt(currentTerm.courtConfigId);
```
bingen: Shouldn't the parameter here be a term id, instead of a court config id?

contracts/court/Court.sol
```diff
@@ -1263,7 +1264,8 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         CourtConfig storage config = _getConfigAtDraftTerm(_round);
         // Court terms are assumed to always fit in uint64. Thus, the end term of a round is assumed to fit in uint64 too.
         // For a regular round, its end term is at the end of its appeal confirmation term.
-        return _round.draftTermId + _round.delayedTerms + config.commitTerms + config.revealTerms + config.appealTerms + config.appealConfirmTerms;
+        uint64 draftTerm = _round.draftTermId + _round.delayedTerms;
```
bingen: Why are calculating this separately?

contracts/court/Court.sol
bingen: Besides, these changes are not related to Solidity 5, right? Then it would be good to have them in a separate PR.

contracts/court/Court.sol
```diff
@@ -1543,13 +1553,13 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     }
 
     // TODO: move to a factory contract
-    function _initSubscriptions(ERC20 _feeToken, uint256[5] _subscriptionParams) private {
-        uint64 MAX_UINT64 = uint64(-1);
-        uint256 MAX_UINT16 = uint16(-1);
+    function _initSubscriptions(ERC20 _feeToken, uint256[5] memory _subscriptionParams) private {
+        uint64 maxUINT64 = uint64(-1);
```
bingen: I guess the linter doesn't complain anymore, but this looks awkward to me. Wouldn't it be more common something like `maxUint64`?

contracts/lib/BytesHelpers.sol
```diff
@@ -1,14 +1,14 @@
-pragma solidity ^0.4.24;
+pragma solidity ^0.5.8;
 
 
 library BytesHelpers {
     function toBytes4(bytes memory _self) internal pure returns (bytes4 result) {
         uint256 length = _self.length;
```
bingen: This seems unnecessary verbose now.

contracts/registry/JurorsRegistry.sol
```diff
@@ -38,7 +38,7 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
     string internal constant ERROR_TOKEN_APPROVE_NOT_ALLOWED = "JR_TOKEN_APPROVE_NOT_ALLOWED";
 
     // Address that will be used to burn juror tokens
-    address internal constant BURN_ACCOUNT = 0xdead;
+    address internal constant BURN_ACCOUNT = address(0x000000000000000000000000000000000000dEaD);
```
bingen: Interesting. Wasn't Solidity 5 accepting the old form?

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -339,8 +339,8 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     }
 
     function _setFeeToken(ERC20 _feeToken) internal {
-        require(isContract(_feeToken), ERROR_NOT_CONTRACT);
-        if (accumulatedGovernorFees > 0) {
+        require(isContract(address(_feeToken)), ERROR_NOT_CONTRACT);
+        if (accumulatedGovernorFees > uint256(0)) {
```
bingen: See this: https://github.com/aragon/aragon-court/pull/103#discussion_r321798681
Let's make sure we are consistent once we make a decision there.

contracts/standards/arbitration/IArbitrable.sol
facuspagnuolo: based on the current `Arbitrable` contract we have

contracts/arbitration/IArbitrable.sol
facuspagnuolo: I prefer using the word `standard` to refer interfaces agreed by the whole community (i.e. EIP). We could rename it to `EIP792` as we are doing for the rest of the standards, but I'm not sure if we are sticking it to the point

contracts/standards/arbitration/IArbitrable.sol
bingen: But we can have an internal function called from the contract wherever is needed and from the external one, like we do in many other places.

contracts/court/Court.sol
facuspagnuolo: > Shouldn't the parameter here be a term id, instead of a court config id?

Yes, I think this was carried from [#103](https://github.com/aragon/aragon-court/pull/103/files?file-filters%5B%5D=.sol#diff-4bb317b489994cf6f4c4b7ed11c18590R850). Nice catch!

> These changes are not related to Solidity 5, right? 

In fact, they are. Note that previously we were declaring a storage pointer to the `CourtConfig` inside the for-loop and using it at the end when transferring founds. This is easy to solve declaring the pointer outside the for-loop, but I was getting a warning due to declaring a storage pointer without initializing it. Therefore, I initialized it to point to the current `CourtConfig`.

> Why do we have 2 calls to `_getConfigSafeAt` now?

I think the explanation above answers this quiestion as well

contracts/court/Court.sol
facuspagnuolo: Answered [here](https://github.com/aragon/aragon-court/pull/109#discussion_r325312162)

contracts/court/Court.sol
facuspagnuolo: To avoid exceeding the max line length while avoiding to have a multi-line math operation

contracts/court/Court.sol
facuspagnuolo: No strong feelings, I can change it for now. However, remember this was a change towards #78

contracts/registry/JurorsRegistry.sol
facuspagnuolo: Nope, it requires properly formed and checksumed addresses now

contracts/subscriptions/CourtSubscriptions.sol
facuspagnuolo: Fixed in https://github.com/aragon/aragon-court/pull/109/commits/a6c49f0b1c1e23a42ca86c558574d290ec2929b3

contracts/lib/BytesHelpers.sol
facuspagnuolo: Fixed in https://github.com/aragon/aragon-court/pull/109/commits/a6c49f0b1c1e23a42ca86c558574d290ec2929b3

contracts/court/Court.sol
facuspagnuolo: Fixed in https://github.com/aragon/aragon-court/pull/109/commits/a6c49f0b1c1e23a42ca86c558574d290ec2929b3

contracts/court/Court.sol
```diff
@@ -223,11 +224,10 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
 
     /**
-    * @dev Ensure the msg.sender is a certain address
-    * @param _addr Address expected to be the msg.sender
+    * @dev Ensure the msg.sender is the CR Voting module
     */
-    modifier only(address _addr) {
-        require(msg.sender == _addr, ERROR_INVALID_SENDER_ADDRESS);
+    modifier onlyVoting() {
+        require(msg.sender == address(voting), ERROR_INVALID_SENDER_ADDRESS);
```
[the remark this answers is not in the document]
facuspagnuolo: Fixed in https://github.com/aragon/aragon-court/pull/109/commits/a6c49f0b1c1e23a42ca86c558574d290ec2929b3

contracts/court/Court.sol
bingen: Oh, I see! Yes, too bad that Solidity was allowing that (we have had similar issues in the past, accidentally using a pointer before its declaration, which was allowed :scream: ).
Thanks for the clarification.

contracts/court/Court.sol
```diff
@@ -851,7 +850,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             emit NewTerm(currentTermId, msg.sender);
 
             // Add amount of fees to be paid for the transitioned term
-            config = _getConfigSafeAt(currentTerm.courtConfigId);
+            config = _getConfigSafeAt(currentTermId);
             totalFee = totalFee.add(config.heartbeatFee.mul(uint256(currentTerm.dependingDrafts)));
```
izqui: Unrelated to this PR: but this is problematic if the Court config changes in between the terms being transitioned and the fee token is changed. We should fix in a separate issue.

facuspagnuolo: Totally, it's weird

facuspagnuolo: Created #127

PR #110, opened 2019-09-03
Fix lifecycle tests

bingen:


facuspagnuolo:
Note that those tests were refactored in #103

PR #113, opened 2019-09-03
Remove redundant condition in hasVotedInFavorOf

bingen:
:bell: Choose either this one or #114

### DIFF
--- a/contracts/CRVoting.sol
+++ b/contracts/CRVoting.sol
@@ -224,8 +224,7 @@ contract CRVoting is Initializable, ICRVoting {
     }
 
     /**
-    * @dev Tell whether a voter voted in favor of a certain outcome in a vote instance or not. If there was no winning
-    *      outcome, it means that no one voted in favor of any of the possible outcomes.
+    * @dev Tell whether a voter voted in favor of a certain outcome in a vote instance or not.
     * @param _voteId ID of the vote instance to query if a voter voted in favor of a certain outcome
     * @param _outcome Outcome to query if the given voter voted in favor of
     * @param _voter Address of the voter to query if voted in favor of the given outcome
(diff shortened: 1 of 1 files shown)


contracts/CRVoting.sol
```diff
@@ -224,16 +224,16 @@ contract CRVoting is Initializable, ICRVoting {
...
+    * @dev Tell whether a voter voted in favor of a certain outcome in a vote instance or not.
     * @param _voteId ID of the vote instance to query if a voter voted in favor of a certain outcome
     * @param _outcome Outcome to query if the given voter voted in favor of
     * @param _voter Address of the voter to query if voted in favor of the given outcome
     * @return True if the given voter voted in favor of the given outcome, false otherwise
     */
     function hasVotedInFavorOf(uint256 _voteId, uint8 _outcome, address _voter) external voteExists(_voteId) view returns (bool) {
         Vote storage vote = voteRecords[_voteId];
-        return vote.winningOutcome != OUTCOME_MISSING && _outcome != OUTCOME_MISSING && vote.votes[_voter].outcome == _outcome;
+        require(_outcome != OUTCOME_MISSING, ERROR_INVALID_OUTCOME);
```
izqui: I wouldn't revert here, IMO it is better to return false and let the consumer (Court) decide what to do.

contracts/CRVoting.sol
```diff
@@ -247,11 +247,12 @@ contract CRVoting is Initializable, ICRVoting {
     */
     function getVotersInFavorOf(uint256 _voteId, uint8 _outcome, address[] _voters) external voteExists(_voteId) view returns (bool[]) {
         Vote storage vote = voteRecords[_voteId];
+        require(_outcome != OUTCOME_MISSING, ERROR_INVALID_OUTCOME);
```
izqui: Reverting here shouldn't be an issue as the Court should never call this with `_outcome = OUTCOME_MISSING`, but I think it is better not to revert anyway

contracts/CRVoting.sol
bingen: I guess you like better #114 then.
The point I don't get: do you mean that if `_outcome` parameter is `OUTCOME_MISSING` and the current outcome of `_voter` is also `OUTCOME_MISSING`, you would return false?
I don't think calling this function wit missing as a parameter makes sense (and it's not called this way anywhere as far as I know), that' why the revert (maybe we could add another function like `hasNotVoted` to check for missing value), but if it does in my opinion it would mean asking "is vote missing for this voter?", which should return true if voter has not voted.

bingen:
Closed in favor of #114.

PR #116, opened 2019-09-04
Revert "checkpointing: simplify binary search algorithm"

bingen:
This reverts commit ee552bf05f6d51f5d0610eeba59979cddf64b0e0.

I'm opening this PR to open up discussion, let's close it if I'm missing something.

What was the rationale for this change? Were any gas measures taken to justify it?

As far as I can tell, binary search right now is only called from `JurorsRegistry.activeBalanceOfAt` and `JurorsRegistry.totalActiveBalanceAt`:

- activeBalanceOfAt: This is only uses `getItemAt` and only for final rounds. I see as very likely that a juror hasn't modified the balance between the draft term and when that function is called.

- totalActiveBalanceAt: This one uses `getTotalAt` for a past term, which in turn seeks for height and root node.
   - Height: This is very very likely to remain unchanged. Actually I don't expect more than 5 levels (maybe 6?), so we may consider to always use backwards linear search (I'll open another PR for that).
   - Node.  Here it would make sense to remove the shortcut.

This function is only used for final rounds too, but called way less times: only once per round vs twice per juror (I think there's a possible optimization here, I'll open another PR for it).

So at a first glance I would say that the shortcut is worth it.


izqui:
We didn't put much thought into it actually (https://github.com/aragon/aragon-court/pull/104#discussion_r317593820), it was a comment mainly motivated by simplifying a little bit.

I think your analysis makes sense, no objection to reverting the change.

PR #117, opened 2019-09-04
Tree: Use always backwards linear search for tree height

bingen:
:bell: Change base to `development` after #115 is merged.

We don't expect more than 5-6 levels of the tree, and the height is
going to change very seldom, the bigger the tree the less
frequent. Therefore we'll be always looking for the last or the
previous to last one value, so it's not worth to use binary tree.


contracts/lib/HexSumTree.sol
```diff
@@ -255,15 +255,6 @@ library HexSumTree {
...
 
-    /**
-    * @dev Tell the height of the tree at a given point in time
-    *      It uses a binary search.
-    * @param _time Point in time to query the height of the tree
-    */
-    function getHeightAt(Tree storage self, uint64 _time) internal view returns (uint256) {
-        return self.height.get(_time);
-    }
-
```
facuspagnuolo: Given that now the only way to access the height of the tree is through the "recent" method, I'd simplify the interface to `getHeightAt`. WDYT?

test/tree/hex-sum-tree-gas.js
```diff
@@ -170,6 +172,13 @@ contract('HexSumTree', () => {
     })
 
     describe('search', () => {
+      const mockNextKeyProgressively = async (exp) => {
+        for (let i = 1; i <= exp; i++) {
+          const nextKey = (new web3.BigNumber(CHILDREN)).pow(i)
```
facuspagnuolo: can we use `bn` here? it was easier to change `BigNumber` only in one place in #108  :)

contracts/test/tree/HexSumTreeGasProfiler.sol
```diff
@@ -56,4 +56,8 @@ contract HexSumTreeGasProfiler is TimeHelpersMock {
     function height() external view returns (uint256) {
         return tree.getHeight();
     }
+
+    function getHeightHistoryLength() external view returns (uint256) {
+        return tree.height.history.length;
+    }
```
facuspagnuolo: Are we using this one somewhere?

contracts/lib/HexSumTree.sol
bingen: Well, I wanted to be consistent with the other methods (like getNode), so, if it has `Recent` in the name it means "linear", otherwise it means "binary" (which makes sense to be the default, as in general is more efficient).

test/tree/hex-sum-tree-gas.js
bingen: Sure, good point! I'll change it.

contracts/test/tree/HexSumTreeGasProfiler.sol
bingen: Nope, sorry. I used it for some gas tests, to make sure that I was testing with enough height history depth, and forgot to remove. I'll do it now, thanks!

Issue #87, opened 2019-08-06
Run unit tests against a geth node


facuspagnuolo:
We should have scripts to run the test suite against a `geth` node as well. 
Explore if we should do this only for certain scenarios in the CI to avoid waiting a lot to merge PRs.

PR #119, opened 2019-09-05
Add scripts to check Court bytecode size

bingen:


bingen:
Yes. I know that tests should fail, but at some point I was not confident about it, although I'm not sure anymore, as I guess that when that happened to me in the past was probably because I was measuring it wrong.
Although I think it does not hurt, we can remove the CI task if you are completely sure we can rely on regular tests and leave only the scripts, which are very useful to check the evolution and how we are doing, as we are often on the bleeding edge.

facuspagnuolo:
I think we can trust our test-suit, EIP-170 is implemented in ganache, and even more, once #87 is implemented, we will be more sure about it. I agree it could be useful to be aware of how far we are of the size limit, but in that case, we should turn those scripts into a github hook or similar, otherwise, I don't think we will remember to run it locally every time we make a change to contracts

bingen:
> we should turn those scripts into a github hook or similar,

Good idea, I'll do that!

facuspagnuolo:
Can we close this one for now?

facuspagnuolo:
Closing this one for now

PR #123, opened 2019-09-10
Chore: Upgrade solidity-coverage to 0.7.0-beta.0

### LINKED ISSUE #86: Add coverage measure tool
We may want to use `solcover` like we are doing in many other repos

cgewecke:
Fixes #86 

@facuspagnuolo Sorry this took a while...

This upgrades to a new solidity-coverage beta which contains breaking changes. It runs on an in-process ganache-core supplied by truffle and launches as a truffle plugin. Instrumentation data is collected in-memory so it should be a bit faster. 

All of the nonsense about `copyPackages`, weird problems with libraries and double-compilation is finished though. Installation is much simpler. Am leaving notes in the diff explaining each change.

Have added myself to the watch list here and please ping if you see anything weird. The beta PR remains open and under active development [here](https://github.com/sc-forks/solidity-coverage/pull/372).

~~[EDIT -- WIP because the `--file <glob>` command option to select a subset of the tests is not grabbing all the files correctly]~~ (Fixed)

[Edit 2 -- 
Unfortunately it looks like there will continue to be things that don't work perfectly  :/

+ **gas measurement tests**: coverage still distorts gas a little and many of the `hex-sum-tree-gas` tests fail. 
+ **very long running individual tests**: In the `registry` there are single tests that take +/- 3 min to complete normally & it looks like coverage is slowing things down another +100%. Those tests timeout, not sure how long they take to complete. 

It might be possible to skip coverage for these without affecting the measurement of logic correctness as such because they're simulations? Idk. 

@facuspagnuolo  Please feel free to close this PR - it's just a draft to explain how the new coverage installs and get a sense of what works (and doesn't). 
]

### DIFF
--- a/.solcover.js
+++ b/.solcover.js
@@ -1,14 +1,6 @@
 module.exports = {
-  norpc: true,
-  compileCommand: '../node_modules/.bin/truffle compile',
-  copyPackages: [
-    '@aragon/os',
-    '@aragon/test-helpers'
-  ],
-  skipFiles: [
-    'test',
-    '@aragon/os',
-    '@aragon/test-helpers',
-  ],
-  deepSkip: true // Turn on deep skip to avoid preprocessing (e.g. removing view/pure modifiers) for skipped files
+  providerOptions: {
+    network_id: 16,
+    total_accounts: 200,
+  }
 }
--- a/.solcover.js
+++ b/.solcover.js
@@ -1,4 +1,7 @@
 module.exports = {
+  skipFiles: [
+    'test',
+  ],
   providerOptions: {
     network_id: 16,
     total_accounts: 200,
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,5 +35,5 @@ jobs:
       script: npm test test/subscriptions/*.js
       name: "Subscriptions"
     - stage: tests
-      script: npx truffle run coverage --file test/registry/*.js
+      script: npx truffle run coverage --file="test/registry/*.js"
       name: "Registry Coverage"
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,5 +35,5 @@ jobs:
       script: npm test test/subscriptions/*.js
       name: "Subscriptions"
     - stage: tests
-      script: npx truffle run coverage --file="test/registry/*.js"
-      name: "Registry Coverage"
+      script: npx truffle run coverage --file="test/voting/*.js"
+      name: "Voting Coverage"
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,3 +34,6 @@ jobs:
     - stage: tests
       script: npm test test/subscriptions/*.js
       name: "Subscriptions"
+    - stage: tests
+      script: npx truffle run coverage --file test/registry/*.js
+      name: "Registry Coverage"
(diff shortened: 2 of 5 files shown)


.solcover.js
```diff
@@ -1,14 +1,6 @@
...
-  skipFiles: [
-    'test',
-    '@aragon/os',
-    '@aragon/test-helpers',
-  ],
-  deepSkip: true // Turn on deep skip to avoid preprocessing (e.g. removing view/pure modifiers) for skipped files
+  providerOptions: {
+    network_id: 16,
+    total_accounts: 200,
+  }
```
cgewecke: We're using an in-process ganache-core - these provider options mirror the one's in your launch script. Deep skip is not an option because SC no longer needs to instrument dependencies. Compilation is handled within a truffle plugin rather than with a shell.exec

.travis.yml
```diff
@@ -34,3 +34,6 @@ jobs:
     - stage: tests
       script: npm test test/subscriptions/*.js
       name: "Subscriptions"
+    - stage: tests
+      script: npx truffle run coverage --file test/registry/*.js
```
cgewecke: Have added a single CI job as an example of how the `--file` command option can be used to partition the coverage as you've done with the tests. You may need [coverall's parallel build feature](https://docs.coveralls.io/parallel-build-webhook) to get this working properly.

test/helpers/assertThrow.js
```diff
@@ -1,5 +1,6 @@
 const REVERT_CODE = 'revert'
-const THROW_ERROR_PREFIX = 'Returned error: VM Exception while processing transaction:'
+const THROW_ERROR_PREFIX_A = 'Returned error: VM Exception while processing transaction:'
+const THROW_ERROR_PREFIX_B = 'VM Exception while processing transaction:' // ganache-core
```
cgewecke: Ganache-core does not include the `Returned error:` portion of the prefix. So...

.travis.yml
cgewecke: This glob is not working as expected - only 198 test of > 700 are running. Need to investigate what's wrong here...
```
test/registry/*.js
```

cgewecke: The glob was auto-expanding somewhere before reaching the coverage tool and only a single item in the file array was received. Fixed by wrapping in quotes and passing to the option e.g:
```
npx truffle run coverage --file="test/registry/*.js"
```

facuspagnuolo:
Thanks for the contribution @cgewecke !! Will merge to another branch and iterate to see how we can workaround those edge cases 👏

cgewecke:
@facuspagnuolo Hi. fwiw will be publishing a new beta over the weekend with things that might be relevant:
 
+ Some speed issues have been addressed
+ Mocha options in solcoverjs so you can grep / invert test cases by test description, e.g
```javascript
it('is a gas usage simulation [ @skip-for-coverage ]', async function(){
...
})
```

```javascript
// .solcover.js
module.exports = {
  mocha: {
    grep: '@skip-for-coverage'
    invert: true
  }
}
```

cgewecke:
New beta at `0.7.0-beta.1`

PR #128, opened 2019-09-19
Chore: Run linter in pre-push hook

facuspagnuolo:
Adding a pre-push githook to run the linter to make sure we don't waste CI jobs unnecessary

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,9 @@
 {
   "name": "@aragon/court",
   "version": "1.0.0",
-  "description": "",
+  "description": "Aragon Court",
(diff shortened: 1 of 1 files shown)


bingen:
I guess I'm missing something, but it's not working for me. Look at this branch I created, I was able to push with linter (which I think was not run) failing:
https://github.com/aragon/aragon-court/tree/ppt

I don't see any `pre-push` created in `.git/hooks`. Maybe we can get rid of this new dependency and just create such a script.

bingen:
Ok, forget it, I deleted node modules and installed again and it worked.

PR #131, opened 2019-09-19
Arbitrable: Remove unused stuff

facuspagnuolo:
This PR includes:
- Removing unused `Agreements` contract
- Removing unused `Arbitrable` contract
- Turn `IArbitrable` into an interface


contracts/arbitration/IArbitrable.sol
```diff
@@ -1,45 +1,46 @@
 pragma solidity ^0.5.8;
 
 
-contract IArbitrable {
+interface IArbitrable {
```
izqui: As we get closer to the end and the `ProposalAgreements` spec is finished, we should update this interface to what Arbitrable actually looks like

PR #134, opened 2019-09-23
Contracts: Minor enhancements

facuspagnuolo:


facuspagnuolo:
> I'm not sure about the rationale for changing error constants to internal.

Was simply to allow subclasses to access them, let's move all them to `private`

PR #135, opened 2019-09-23
Subscriptions: Write inline documentation

### LINKED ISSUE #79: Contracts: Write inline documentation
Add proper documentation for all the contracts of the court:
- Interface documentation (function, parameters, return values)
- Radspec description when needed
- Clarify non-trivial logic 

Contracts that MUST be documented
- [x] Main Court contract (https://github.com/aragon/aragon-court/pull/103)
- [x] Jurors registry (https://github.com/aragon/aragon-court/pull/72)
- [x] CR Voting (https://github.com/aragon/aragon-court/pull/96)
- [x] HexSumTree (https://github.com/aragon/aragon-court/pull/104)
- [x] Checkpointing (https://github.com/aragon/aragon-court/pull/104)
- [x] Accounting (https://github.com/aragon/aragon-court/pull/67)
- [ ] Subscriptions (https://github.com/aragon/aragon-court/pull/135)

facuspagnuolo:
**MERGE AFTER #134**

Fixes #79

### DIFF
--- a/contracts/subscriptions/CourtSubscriptions.sol
+++ b/contracts/subscriptions/CourtSubscriptions.sol
@@ -17,7 +17,8 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     using SafeMath for uint256;
     using PctHelpers for uint256;
 
-    uint64 internal constant START_TERM_ID = 1; // term 0 is for jurors onboarding
+    // Term 0 is for jurors on-boarding
+    uint64 internal constant START_TERM_ID = 1;
 
     string internal constant ERROR_NOT_GOVERNOR = "SUB_NOT_GOVERNOR";
     string internal constant ERROR_OWNER_ALREADY_SET = "SUB_OWNER_ALREADY_SET";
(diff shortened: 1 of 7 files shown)


contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -104,18 +134,10 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     }
 
     /**
-     * @notice Pay fees on behalf of `_from` for `_periods` periods
-     * @param _from Subscriber whose subscription is being paid
-     * @param _periods Number of periods to be paid in total (delayed plus regular)
-     * @dev This is a graphical explanation of a generic case:
```
bingen: I liked this better where it was.

facuspagnuolo: I ended up changing it again in #138, let me know if you like the last change or if you still prefer having it here. Will merge and update it #138 to avoid conflicts.

facuspagnuolo:
> I see again here the `private` -> `internal` change for error messages

I think GitHub was showing an out-dated diff... I tried merging it locally and error constants remain  `private` in `development` 🤷‍♂

contracts/subscriptions/CourtSubscriptions.sol
bingen: Well, at least now it's in a header, but the point is that the diagram was trying to explain the meaning of `_periods` param in `payFees` function.
If you keep above `_getDelayedPeriods`, at least I think `_periods` variable should be changed, as it doesn't make so much sense there (like maybe remove the underscore)

PR #136, opened 2019-09-23
Subscriptions: Organize functions according linter rules

facuspagnuolo:
**MERGE AFTER https://github.com/aragon/aragon-court/pull/135**

### DIFF
--- a/contracts/subscriptions/CourtSubscriptions.sol
+++ b/contracts/subscriptions/CourtSubscriptions.sol
@@ -17,9 +17,6 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     using SafeMath for uint256;
     using PctHelpers for uint256;
 
-    // Term 0 is for jurors on-boarding
-    uint64 internal constant START_TERM_ID = 1;
-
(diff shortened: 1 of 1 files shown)


contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -359,6 +354,46 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
         emit GovernorFeesTransferred(amount);
     }
 
+    /**
+    * @dev Internal function to make sure the fee token address and amount of a certain period have been cached
+    * @param _period Period being ensured to have cached its fee token address and amount
+    * @return feeToken ERC20 token to be used for the subscription fees during the given period
+    * @return feeAmount Amount of fees to be paid during the given period
+    */
+    function _ensurePeriodFeeTokenAndAmount(Period storage _period) internal returns (ERC20 feeToken, uint256 feeAmount) {
```
bingen: Just curious: which linter rules are this one and next one (`_ensurePeriodBalanceDetails`) breaking?

facuspagnuolo: None, just placing them above to be consistent with the order I used for the external interface

PR #137, opened 2019-09-23
Subscriptions: Polish unit tests

facuspagnuolo:
**MERGE AFTER #136**

Fixes #81

PR #138, opened 2019-09-23
Subscriptions: Polish contract

facuspagnuolo:
**MERGE AFTER #137**

### DIFF
--- a/contracts/subscriptions/CourtSubscriptions.sol
+++ b/contracts/subscriptions/CourtSubscriptions.sol
@@ -227,6 +227,7 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     * @param _feeAmount New amount of fees to be paid for each subscription period
     */
     function setFeeAmount(uint256 _feeAmount) external onlyGovernor {
+        require(currentFeeAmount != _feeAmount, ERROR_INVALID_FEE_AMOUNT);
         _setFeeAmount(_feeAmount);
     }
 
@@ -237,8 +238,10 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     * @param _feeAmount New amount of fees to be paid for each subscription period
     */
     function setFeeToken(ERC20 _feeToken, uint256 _feeAmount) external onlyGovernor {
-        // `setFeeToken` transfers governor's accumulated fees, so must be executed first
+        require(currentFeeToken != _feeToken, ERROR_INVALID_FEE_TOKEN);
+
         _setFeeToken(_feeToken);
+        // `setFeeToken` transfers governor's accumulated fees, so must be executed first
         if (currentFeeAmount != _feeAmount) {
             _setFeeAmount(_feeAmount);
         }
@@ -249,6 +252,7 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     * @param _prePaymentPeriods New number of periods that can be paid in advance
     */
     function setPrePaymentPeriods(uint256 _prePaymentPeriods) external onlyGovernor {
+        require(prePaymentPeriods != _prePaymentPeriods, ERROR_INVALID_PREPAYMENT_PERIODS);
         _setPrePaymentPeriods(_prePaymentPeriods);
     }
 
@@ -257,6 +261,7 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     * @param _latePaymentPenaltyPct New per ten thousand of subscription fees that will be applied as penalty for not paying during proper period
     */
     function setLatePaymentPenaltyPct(uint16 _latePaymentPenaltyPct) external onlyGovernor {
+        require(latePaymentPenaltyPct != _latePaymentPenaltyPct, ERROR_INVALID_LATE_PAYMENT_PENALTY_PCT);
         _setLatePaymentPenaltyPct(_latePaymentPenaltyPct);
     }
 
@@ -265,6 +270,7 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     * @param _governorSharePct New per ten thousand of subscription fees that will be allocated to the governor of the Court (‱ - 1/10,000)
     */
     function setGovernorSharePct(uint16 _governorSharePct) external onlyGovernor {
+        require(governorSharePct != _governorSharePct, ERROR_INVALID_GOVERNOR_SHARE_PCT);
         _setGovernorSharePct(_governorSharePct);
     }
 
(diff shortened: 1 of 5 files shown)


contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -121,15 +131,15 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
...
-        require(_periodDuration > 0, ERROR_ZERO_PERIOD_DURATION);
+        require(_periodDuration > 0, ERROR_PERIOD_DURATION_ZERO);
 
         owner = _owner;
         jurorsRegistry = _jurorsRegistry;
         periodDuration = _periodDuration;
         _setFeeToken(_feeToken);
         _setFeeAmount(_feeAmount);
         _setPrePaymentPeriods(_prePaymentPeriods);
-        latePaymentPenaltyPct = _latePaymentPenaltyPct;
```
bingen: I don't get this. What's the point of converting a simple assignment into a function? I would say this is quite atomic, simple and readable. Are we supposed to convert all assignments into functions?
Besides, as far as I know Solidity doesn't have `inline` functions (modifiers a part).

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -221,7 +239,9 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     function setFeeToken(ERC20 _feeToken, uint256 _feeAmount) external onlyGovernor {
         // `setFeeToken` transfers governor's accumulated fees, so must be executed first
         _setFeeToken(_feeToken);
-        _setFeeAmount(_feeAmount);
+        if (currentFeeAmount != _feeAmount) {
```
bingen: :+1:

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -414,43 +421,71 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
...
     * @dev Internal function to set a new amount for the subscription fees
-    * @param _feeAmount New amount of fees to be paid for each subscription period
+    * @param _newFeeAmount New amount of fees to be paid for each subscription period
     */
-    function _setFeeAmount(uint256 _feeAmount) internal {
-        require(_feeAmount > 0, ERROR_ZERO_FEE);
-        currentFeeAmount = _feeAmount;
+    function _setFeeAmount(uint256 _newFeeAmount) internal {
+        uint256 previousFeeAmount = currentFeeAmount;
+        require(previousFeeAmount != _newFeeAmount, ERROR_INVALID_FEE_AMOUNT);
```
bingen: This sounds like too much overhead to me, unnecessary complexity and gas waste. What's the problem if somebody tries to set the amount to the same one?
Same applies to other setters below.

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -469,7 +504,7 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
     */
     function _getPeriodStartTermId(uint256 _periodId) internal view returns (uint64) {
         // Periods are measured in Court terms. Since Court terms are represented in `uint64`, we are safe to use `uint64` for period ids too.
-        return START_TERM_ID + uint64(_periodId) * periodDuration;
+        return START_TERM_ID + uint64(_periodId * periodDuration);
```
bingen: Why this change? `periodDuration` is already an `uint64`.

contracts/subscriptions/CourtSubscriptions.sol
facuspagnuolo: Please, review the setter first

contracts/subscriptions/CourtSubscriptions.sol
facuspagnuolo: Emitting duplicated events, I'd always prioritize usability vs gas consumption, even more if we are talking about non-daily-basis user functionality like this function

contracts/subscriptions/CourtSubscriptions.sol
facuspagnuolo: True, not sure what I meant here, maybe it was a typo 🤷‍♂

contracts/subscriptions/CourtSubscriptions.sol
bingen: Hm, ok, sounds reasonable.

contracts/subscriptions/CourtSubscriptions.sol
bingen: Good point, I was assuming it would do the same. Still, I don't think we need all that extra stuff. We could discuss about the event (do we  need it on `init`?), but right now it would be impossible to init the contract wit a `latePaymentPenaltyPct = 0`, right? I don't think this is what we want.

facuspagnuolo: True, updated the internal setters to avoid checking diffs

bingen: I see you removed the check but left the function and the event. I still don't think we need the event on initialization (and therefore the function). Not sure, as this is more a front-end issue, but we are not doing this in aragon-apps. See some examples:
- Finance:
https://github.com/aragon/aragon-apps/blob/master/apps/finance/contracts/Finance.sol#L177
https://github.com/aragon/aragon-apps/blob/master/apps/finance/contracts/Finance.sol#L311

- Voting:
https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L86
https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L103
https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L117

contracts/subscriptions/CourtSubscriptions.sol
bingen: Altough btw we don't do this on aragon-apps either

https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L95
https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L110
https://github.com/aragon/aragon-apps/blob/master/apps/finance/contracts/Finance.sol#L304

We could bring it up there. Pinging @sohkai to hear his opinion about it.

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -514,31 +543,53 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
...
+    * @param _subscriber Subscriber querying the delayed periods of
+    * @param _currentPeriodId Identification number of the current period
+    * @return Number of overdue payments for the requested subscriber
+    *
+    *    subs      last           cur       new
+    *   +----+----+----+----+----+----+----+----+
+    *                  <---------><------------->
+    *                    delayed      regular
+    *                  <------------------------>
+    *                          _periods
```
bingen: Quick reminder: either move it back to `payFees` function or change this `_periods` variable that doesn't belong to here.

contracts/subscriptions/CourtSubscriptions.sol
bingen: I've been thinking more about it and I'm again not convinced with this change. This can be easily prevented in the front-end. I remember once @izqui telling me that our job is not to make front-end life easy :wink: and I guess that's why we don't so these things in aragon-apps. I guess we should keep that philosophy.

facuspagnuolo: > Altough btw we don't do this on aragon-apps either

Nice, a few more things to fix then :)

> This can be easily prevented in the front-end.

Everything can be prevented by the front-end

Seriously, no strong feelings, emitting "falsey" events is not a good practice IMO. This is sth I'd expect from an auditor's feedback for example. However, I don't want to waste time on something non-critical, I'll roll back this behavior.

bingen: > Everything can be prevented by the front-end

Haha, fair enough, but you know what I meant right? That as you said this is not critical.

> emitting "falsey" events is not a good practice IMO

Yes, as I said I see your points... although prevent "falsey" events would be as simple as renaming them from `...Changed` to `...Set` :stuck_out_tongue_winking_eye: 

> I'll roll back this behavior

Sounds good, although maybe we can wait for @izqui or @sohkai to confirm that this is a common and intended practice of ours. Or maybe we can just cherry-pick this if we decide we want it.

bingen: Just for the record, I want to clarify this a little bit more for further discussion (as "not critical" sounds too vague)

>>> I'd always prioritize usability vs gas consumption
>> This can be easily prevented in the front-end.
> Everything can be prevented by the front-end

The way I see we have been working so far is: security (and gas consumption) must be taken care of in smart contracts, usability belongs to front-end (of course there are exceptions, like bundling two transactions in one)

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -514,31 +535,51 @@ contract CourtSubscriptions is IsContract, ISubscriptions, TimeHelpers {
...
+    /**
+    * @dev Internal function to tell the number of overdue payments for a given subscriber
+    * @param _subscriber Subscriber querying the delayed periods of
+    * @param _currentPeriodId Identification number of the current period
+    * @return Number of overdue payments for the requested subscriber
+    *
+    *    subs      last           cur       new
+    *   +----+----+----+----+----+----+----+----+
+    *                  <---------><------------->
+    *                    delayed      regular
```
bingen: To be honest, I don't think this makes much sense the way it is here. This was intended to explain the whole picture, how the paid periods (this line has now been removed) were the sum of `delayed` plus `regular` and how it related to timeline. So the `regular` part doesn't apply here.
I'd rather remove it.

PR #139, opened 2019-09-23
Contracts: Use Initializable contract

facuspagnuolo:
**MERGE AFTER #138**


contracts/court/CourtAccounting.sol
```diff
@@ -30,7 +30,10 @@ contract CourtAccounting is IAccounting {
     }
 
     function init(address _owner) external {
-        require(owner == address(0), ERROR_ALREADY_INITIALIZED);
```
bingen: To be honest, it looks like a lot of overhead for something that can be done with one line, but not opposed to it.

Issue #102, opened 2019-08-12
Registry: Add jurors' active balance limit validation


facuspagnuolo:
Based on how we compute jurors weight for final appeal rounds:
```
uint256 weight = FINAL_ROUND_WEIGHT_PRECISION.mul(activeBalance / minJurorsActiveBalance)
```

It means that jurors' active balance cannot be greater than `2^64 * minJurorsActiveBalance / FINAL_ROUND_WEIGHT_PRECISION` or simply `2^64 / FINAL_ROUND_WEIGHT_PRECISION` assuming the smallest `minJurorsActiveBalance`. We should add a validation in the activation process to ensure that holds.

PR #142, opened 2019-09-24
Registry: Implement total active balance limit

facuspagnuolo:
**MERGE AFTER #141**

Fixes https://github.com/aragon/aragon-court/issues/102


contracts/registry/JurorsRegistry.sol
```diff
@@ -110,16 +113,19 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
...
         // TODO: cannot check the given owner is a contract cause the Court set this up in the constructor, move to a factory
         // require(isContract(_owner), ERROR_NOT_CONTRACT);
         require(isContract(address(_jurorToken)), ERROR_NOT_CONTRACT);
+        require(_minActiveBalance.mul(_minActiveBalancesLimit) > 0, ERROR_BAD_MIN_ACTIVE_BALANCES_LIMIT);
 
         initialized();
         owner = _owner;
         jurorsToken = _jurorToken;
         minActiveBalance = _minActiveBalance;
+        minActiveBalancesLimit = _minActiveBalancesLimit;
```
bingen: Instead of storing this, I would store what you have below:
```
uint256 maxTotalActiveBalance = minActiveBalance * minActiveBalancesLimit;
```
This way you don't need to do the math each time a balance is updated (and the function `_checkTotalActiveBalance` is easier to understand IMO).

facuspagnuolo:
Thanks @bingen for the review, super helpful! I already addressed all your comments, let me know your thoughts!

PR #145, opened 2019-09-25
Registry: Take into account deactivation requests when drafting

### LINKED ISSUE #100: Registry: Handle jurors deactivation request when drafting
Currently, we are not checking if we have to reduce a deactivation request when we draft jurors. If a juror already requested a deactivation amount for the next term and he gets drafted in the current term, he may not have enough balance in the next term to process the submitted deactivation request. We should do something similar to what we do in `collectTokens`

facuspagnuolo:
Fixes #100 

This PR includes:
- Excluding jurors for a draft if they won't have enough active balance based on a future deactivation request
- Small optimization when drafting


contracts/registry/JurorsRegistry.sol
```diff
@@ -225,64 +236,25 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
...
-                sortitionIteration
-            ];
-
-            (uint256[] memory jurorIds, uint256[] memory activeBalances) = _treeSearch(treeSearchParams);
-            for (uint256 i = 0; i < jurorIds.length; i++) {
-                // We assume the selected jurors are registered in the registry, we are not checking their addresses exist
-                address juror = jurorsAddressById[jurorIds[i]];
-                Juror storage _juror = jurorsByAddress[juror];
-                uint256 newLockedBalance = _juror.lockedBalance.add(_draftLockAmount(uint16(_params[6]))); // _penaltyPct
+        DraftParams memory draftParams = DraftParams({
```
izqui: Nit: How about moving this to an internal function that takes the `_params` array and returns a `DraftParams`?

contracts/registry/JurorsRegistry.sol
```diff
@@ -347,8 +319,9 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
             return false;
         }
 
-        // Check if the amount of active tokens is enough to collect the requested amount,
-        // otherwise reduce the requested deactivation amount of the next term
+        // Check if the amount of active tokens is enough to collect the requested amount, otherwise reduce the requested deactivation amount of
+        // the next term. Note that this behaviour is different to the one when drafting jurors since this function is called deliberately by
```
izqui:
```suggestion
        // the next term. Note that this behaviour is different to the one when drafting jurors since this function is called as a side effect of a juror deliberately voting in a final round
```

contracts/registry/JurorsRegistry.sol
```diff
@@ -664,6 +637,57 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
...
+                // We assume the selected jurors are registered in the registry, we are not checking their addresses exist
+                address juror = jurorsAddressById[jurorIds[i]];
+                Juror storage _juror = jurorsByAddress[juror];
+
+                // Compute new locked balance for a juror based on the penalty applied when being drafted
+                uint256 newLockedBalance = _juror.lockedBalance.add(_draftParams.draftLockAmount);
+
+                // Check if there is any deactivation requests for the next term. Drafts are always computed for the current term
+                // but we have to make sure we are locking an amount that will exist in the next term.
+                uint256 nextTermDeactivationRequestAmount = _deactivationRequestedAmountForTerm(_juror, _draftParams.termId + 1);
```
izqui: Just a sanity check for me, it is only possible to have a deactivation request for the very next term, right?

facuspagnuolo: Yup

contracts/registry/JurorsRegistry.sol
```diff
@@ -225,64 +236,14 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
...
-                    selectedJurors++;
-
-                    //                _disputeId
-                    emit JurorDrafted(_params[1], juror);
-                    batchRequestedJurors--;
-                }
-            }
-        }
+        DraftParams memory draftParams = _buildDraftParams(_params);
+        length = _draft(draftParams, jurors, weights);
```
bingen: What's the point of moving this to a separate function? It's the core logic of this function, it isn't reused anywhere and (IMO) it makes it harder to follow the logic (you have to jump to the other function, remember what params are...)
Besides, I don't like the fact that inside that function we are using the params as mutable inner variables.

contracts/registry/JurorsRegistry.sol
```diff
@@ -225,64 +236,14 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
     *
     * @return jurors List of jurors selected for the draft
     * @return weights List of weights corresponding to each juror
-    * @return outputLength Size of the list of the draft result
-    * @return selectedJurors Number of seats selected for the draft
+    * @return length Size of the list of the draft result
     */
-    function draft(uint256[7] calldata _params) external onlyOwner
-        returns (address[] memory jurors, uint64[] memory weights, uint256 outputLength, uint64 selectedJurors)
```
bingen: I think we should keep returning `selectedJurors` and allow the function to return a value less than `_selectedJurors + _requestedJurors`, to prevent this:
```
// Note that this loop could end with an OOG error.
```
So I would add a hard cap to `sortitionIteration` so the inner loop can end before finding all requested jurors.

contracts/registry/JurorsRegistry.sol
facuspagnuolo: > What's the point of moving this to a separate function?

To me it makes a lot of sense, we do it every time, we extract stuff to other functions to avoid having large functions. 

> it makes it harder to follow the logic

It is a 5-lines function where the first 4 lines are parsing the given input, and the last one calling the inner function. 

> Besides, I don't like the fact that inside that function we are using the params as mutable inner variables.

Not sure which case you're specifically talking about, it is a language constraint until we decide to use ABI encoder V2 or support structs natively

contracts/registry/JurorsRegistry.sol
facuspagnuolo: I agree about the cap. However, I still think we may not need to return the `selectedJurors` since it could be handled with the array. Would you use a hardcoded constant for the cap? Let's say using a delta above the number of jurors to be found for example?

facuspagnuolo: I've been discussing the idea of limiting the number of iterations with @izqui and I think it requires a bit more analysis since there are some consequences that are not trivial. Opened [this issue](https://github.com/aragon/aragon-court/issues/157) to tackle this problem

Issue #126, opened 2019-09-18
Court: Turn disputes initial jurors number into a setting


facuspagnuolo:
Currently, every time a user creates a dispute, the number of jurors to be drafted has to be given. We should have the initial number of jurors as part of the `CourtSettings`

facuspagnuolo:
Related to https://github.com/aragon/aragon-court/issues/106

facuspagnuolo:
Closed by https://github.com/aragon/aragon-court/pull/146

Issue #125, opened 2019-09-18
Court: Move appeal factors to court settings


facuspagnuolo:
Move `Court` constants `APPEAL_COLLATERAL_FACTOR` and `APPEAL_CONFIRMATION_COLLATERAL_FACTOR` to `CourtSettings`

facuspagnuolo:
Closed by https://github.com/aragon/aragon-court/pull/146

bingen:
TIL: github's grammar knowledge is limited, so if you say "Fixes #106, #125 and #126", it will only fix the first one, but the other two will remain open. :joy: 
Thanks @facuspagnuolo for doing it for me! :wink:

Issue #127, opened 2019-09-19
Court: Limit the heartbeat transitions up to the next court config if scheduled


facuspagnuolo:
If there is an upcoming court config already scheduled, we should limit the max number of transitions for the heartbeat function to avoid having weird situations like changing the fee token.

Issue #90, opened 2019-08-06
Court: Improve court settings change flow


facuspagnuolo:
Currently, the `Court` settings are updated simply for a given term id. This allows some weird scenarios such as:
- Changing the court settings during a dispute (different rounds could use different settings)
- Having fees and term settings together, it may be interesting if it could help to have different flows to update each of them
- Jurors will have to face different periods for each case if their corresponding settings get updated.

PR #146, opened 2019-09-25
Court: Implement new config params

### LINKED ISSUE #106: Court: bounds for initial juror number when creating a dispute
There is currently a to-do for limiting the max number of initial jurors, but we should also require a minimum of 1 juror for creating the dispute. A dispute with 0 jurors would revert when drafting, and even if it probably doesn't create an issue for the Court outside of this dispute, we should prevent it from happening in the first place.

Setting a very high number of initial jurors would make appeals really expensive, as the total number of appeal rounds before getting to the final round is fixed.

Another option is to make the initial number of jurors a configuration parameter, so creators of a dispute don't have the ability to decide. A counter argument to this would be that having the option to set the initial juror number allows the creator of a dispute to start with more or less 'security' and more certainty that the dispute will be finalized sooner.

bingen:
Move initial number of jurors of disputes and appeal collateral params to config.

Several EVM limitations were hit again:
- Stack too deep forced to split config in 2 different structs
- Which in turn forced to manually write a getter
- Deployment tx gas is now over 6.9e6

Fixes #106, #125 and #126


contracts/court/Court.sol
```diff
@@ -111,67 +106,78 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+        uint64 revealTerms;            // Revealing period duration in terms
+        uint64 appealTerms;            // Appealing period duration in terms
+        uint64 appealConfirmTerms;     // Confirmation appeal period duration in terms
+        uint16 penaltyPct;             // Per ten thousand that will be used to compute the tokens to be locked for a juror on a draft
+        uint16 finalRoundReduction;    // ‱ of reduction applied for final appeal round (1/10,000)
+        uint16 firstRoundJurorsNumber; // Number of jurors drafted on first round
+        uint16 appealStepFactor;       // Factor in which the jurors number is increased on each appeal
+        uint16 maxRegularAppealRounds; // Before the final appeal
+        uint8 appealCollateralFactor;  // Multiple of juror fees required to appeal a preliminary ruling
+        uint8 appealConfirmCollateralFactor; // Multiple of juror fees required to confirm appeal
```
facuspagnuolo: We could use `uint256` for some of these variables avoiding to cast every time we need to operate with these values. For instance `appealCollateralFactor`, `appealConfirmCollateralFactor`, and `maxRegularAppealRounds`.

Additionally `firstRoundJurorsNumber ` and `appealStepFactor` could be a `uint64` based on how we are using them.

contracts/court/Court.sol
```diff
@@ -334,11 +347,10 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     * @dev Create a dispute to be drafted in a future term
     * @param _subject Arbitrable subject being disputed
     * @param _possibleRulings Number of possible rulings allowed for the drafted jurors to vote on the dispute
-    * @param _jurorsNumber Requested number of jurors to be drafted for the dispute
```
facuspagnuolo: I cannot link the `@notice` doc, but it has to be updated as well.

contracts/court/Court.sol
```diff
@@ -728,6 +741,43 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         return governor;
     }
 
+    /**
+     * @dev Get Court configuration parameters
+     * @return token Address of the token used to pay for fees
+     * @return roundStateDurations Array containing number of terms that the different states a dispute round last,
```
facuspagnuolo:
```suggestion
     * @return roundStateDurations Array containing the durations in terms of the different phases a dispute goes through,
```

contracts/court/Court.sol
```diff
@@ -271,10 +277,18 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     * @param _governor Address of the governor contract.
     * @param _firstTermStartTime Timestamp in seconds when the court will open (to give time for juror onboarding)
     * @param _minJurorsActiveBalance Minimum amount of juror tokens that can be activated
-    * @param _roundStateDurations Number of terms that the different states a dispute round last
+    * @param _roundStateDurations Number of terms that the different states a dispute round last,
```
facuspagnuolo:
```suggestion
    * @param _roundStateDurations Array containing the durations in terms of the different phases of a dispute,
```

contracts/court/Court.sol
```diff
@@ -1142,10 +1191,16 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         require(configChangeTermId > termId || termId == ZERO_TERM_ID, ERROR_PAST_TERM_FEE_CHANGE);
 
         // Make sure the given penalty pct is not greater than 100%
+        uint16 _penaltyPct = _pcts[0];
         require(PctHelpers.isValid(_penaltyPct), ERROR_INVALID_PENALTY_PCT);
 
+        // A dispute with 0 jurors in the first round wouldn't make any sense
```
facuspagnuolo:
```suggestion
        // Disputes must request at least one juror to be drafted initially
```

contracts/court/Court.sol
```diff
@@ -29,6 +29,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     string private constant ERROR_BAD_FIRST_TERM_START_TIME = "CT_BAD_FIRST_TERM_START_TIME";
     string private constant ERROR_CONFIG_PERIOD_ZERO_TERMS = "CT_CONFIG_PERIOD_0";
     string private constant ERROR_INVALID_PENALTY_PCT = "CT_INVALID_PENALTY_PCT";
+    string private constant ERROR_BAD_INITIAL_JURORS = "CT_ERROR_BAD_INITIAL_JURORS";
```
facuspagnuolo: WDYT about being explicit here renaming it to `ERROR_BAD_INITIAL_JURORS_NUMBER`

contracts/court/Court.sol
```diff
@@ -728,6 +741,43 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+     *         penaltyPct ‱ of minJurorsActiveBalance that can be slashed (1/10,000)
+     *         finalRoundReduction ‱ of fee reduction for the last appeal round (1/10,000)
+     * @return roundParams Array containing params for rounds:
+     *         firstRoundJurorsNumber Number of jurors to be drafted for the first round of disputes
+     *         appealStepFactor Increasing factor for the number of jurors of each round of a dispute
+     *         maxRegularAppealRounds Number of regular appeal rounds before the final round is triggered
+     * @return appealCollateralParams Array containing params for appeal collateral:
+     *         appealCollateralFactor Multiple of juror fees required to appeal a preliminary ruling
+     *         appealConfirmCollateralFactor Multiple of juror fees required to confirm appeal
+     */
```
facuspagnuolo: Nit: Note that I applied the other doc indentation everywhere:
```
/**
*
```

instead of 

```
/**
 *
```

contracts/court/Court.sol
```diff
@@ -1308,8 +1373,18 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    }
+
+    /**
+     * @dev Internal function to calculate the jurors number for the first regular round of a dispute.
+     * @param _draftTermId Term from which the the jurors for the dispute will be drafted
+     * @return Jurors number for the first regular round of a dispute
+     */
+    function _getFirstRoundJurorsNumber(uint64 _draftTermId) internal view returns (uint64 jurorsNumber) {
+        CourtConfig storage config = _getConfigAt(_draftTermId);
+        jurorsNumber = config.disputes.firstRoundJurorsNumber;
```
facuspagnuolo: Can we avoid the non-explicit return here?

contracts/court/Court.sol
```diff
@@ -1308,8 +1373,18 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
-        confirmAppealDeposit = totalFees * APPEAL_CONFIRMATION_COLLATERAL_FACTOR;
+        appealDeposit = totalFees * config.disputes.appealCollateralFactor;
+        confirmAppealDeposit = totalFees * config.disputes.appealConfirmCollateralFactor;
+    }
+
+    /**
+     * @dev Internal function to calculate the jurors number for the first regular round of a dispute.
+     * @param _draftTermId Term from which the the jurors for the dispute will be drafted
+     * @return Jurors number for the first regular round of a dispute
+     */
```
facuspagnuolo: Nit: inline doc indentation

test/court/court-appeal.js
```diff
@@ -20,23 +20,24 @@ contract('Court', ([_, disputer, drafter, appealMaker, appealTaker, juror500, ju
     { address: juror3500, initialActiveBalance: bigExp(3500, 18) },
     { address: juror2500, initialActiveBalance: bigExp(2500, 18) },
   ]
+  const firstRoundJurorsNumber = 3
 
   beforeEach('create court', async () => {
     courtHelper = buildHelper()
-    court = await courtHelper.deploy()
+    court = await courtHelper.deploy({ firstRoundJurorsNumber })
```
facuspagnuolo: I think we can rely on the default config for tests unless we need to change it

test/court/court-appeal.js
```diff
@@ -162,9 +163,9 @@ contract('Court', ([_, disputer, drafter, appealMaker, appealTaker, juror500, ju
...
                     assert.equal(draftTerm.toString(), draftTermId, 'current round draft term does not match')
                     assert.equal(delayedTerms.toString(), 0, 'current round delay term does not match')
-                    assert.equal(roundJurorsNumber.toString(), jurorsNumber, 'current round jurors number does not match')
-                    assert.equal(selectedJurors.toString(), jurorsNumber, 'current round selected jurors number does not match')
-                    assert.equal(jurorFees.toString(), courtHelper.jurorFee.mul(bn(jurorsNumber)).toString(), 'current round juror fees do not match')
+                    assert.equal(roundJurorsNumber.toString(), firstRoundJurorsNumber, 'current round jurors number does not match')
+                    assert.equal(selectedJurors.toString(), firstRoundJurorsNumber, 'current round selected jurors number does not match')
+                    assert.equal(jurorFees.toString(), courtHelper.jurorFee.mul(bn(firstRoundJurorsNumber)).toString(), 'current round juror fees do not match')
```
facuspagnuolo: Note there is `DEFAULTS` object being exported from the court helper module, we should use that one instead of defining a new variable for it (`firstRoundJurorsNumber`)

test/court/court-config.js
```diff
@@ -0,0 +1,76 @@
...
+      assertBn(roundStateDurations[2], appealTerms, 'Appeal terms number does not match')
+      assertBn(roundStateDurations[3], appealConfirmTerms, 'Appeal confirmation terms number does not match')
+      assertBn(pcts[0], penaltyPct, 'Penalty permyriad does not match')
+      assertBn(pcts[1], finalRoundReduction, 'Final round reduction does not match')
+      assertBn(roundParams[0], firstRoundJurorsNumber, 'First round jurors number does not match')
+      assertBn(roundParams[1], appealStepFactor, 'Appeal step factor does not match')
+      assertBn(roundParams[2], maxRegularAppealRounds, 'Number af max regular appeal rounds does not match')
+      assertBn(appealCollateralParams[0], appealCollateralFactor, 'Appeal collateral factor does not match')
+      assertBn(appealCollateralParams[1], appealConfirmCollateralFactor, 'Appeal confirmation collateral factor does not match')
+    })
```
facuspagnuolo: Should we place this one in `court-initialize.js`?

test/court/court-initialize.js
```diff
@@ -26,13 +26,15 @@ contract('Court', () => {
...
+        const termId = await court.getLastEnsuredTermId()
+        const penaltyPct = (await court.getCourtConfig(termId)).pcts[0]
         assert.equal(penaltyPct.toString(), 0, 'penalty pct does not match')
       })
 
       it('can be 100%', async () => {
         const court = await courtHelper.deploy({ penaltyPct: bn(10000) })
-        const penaltyPct = (await court.courtConfigs(1))[9] // config ID 0 is used for undefined
+        const termId = await court.getLastEnsuredTermId()
+        const penaltyPct = (await court.getCourtConfig(termId)).pcts[0]
```
facuspagnuolo: We can define a new getter in the court helper module that provides the court settings properly parsed :)

test/helpers/numbers.js
```diff
@@ -9,11 +9,16 @@ const MAX_UINT64 = maxUint(64)
...
+}
+
 module.exports = {
   bn,
   bigExp,
   ONE,
   MAX_UINT64,
   MAX_UINT192,
   MAX_UINT256,
+  assertBn,
```
facuspagnuolo: 👍

test/court/court-draft.js
```diff
@@ -18,10 +18,11 @@ contract('Court', ([_, disputer, drafter, juror500, juror1000, juror1500, juror2
     { address: juror1500, initialActiveBalance: bigExp(1500, 18) },
     { address: juror2000, initialActiveBalance: bigExp(2000, 18) },
   ]
+  const jurorsNumber = 5
 
   beforeEach('create court', async () => {
     courtHelper = buildHelper()
-    court = await courtHelper.deploy()
+    court = await courtHelper.deploy({ jurorsNumber })
```
facuspagnuolo: I think you meant `firstRoundJurorsNumber`

bingen:
Hey @facuspagnuolo wait a second please, I didn't assign you as reviewer yet, because it's not finished yet (you'll see several force-pushes).
I'll let you know when I'm done (just trying not to waste your time :wink: ) 
Anyway thanks for your comments!

contracts/court/Court.sol
bingen: Yes, but it [used to be the other way around](https://github.com/aragon/aragon-court/blob/d5356bc2e45a5151945346c0ee3398b43b882bb1/contracts/Court.sol#L172) (I don't know why it was changed), which is a quite common practice, see for instance: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#commenting
Please, let's write our style guide before getting into this unfruitful discussions.

facuspagnuolo: I actually like better the way you did, but wanted to be consistent with what we have in apps, maybe there are other repos using your style 🤷‍♂

bingen: Well, we have a mix, see for instance:
https://github.com/aragon/aragon-apps/blob/master/apps/agent/contracts/Agent.sol#L251
https://github.com/aragon/aragonOS/blob/next/contracts/apps/AppProxyUpgradeable.sol#L20
https://github.com/aragon/aragonOS/blob/next/contracts/lib/ens/ENS.sol#L33
https://github.com/aragon/aragonOS/blob/next/contracts/lib/math/Math.sol#L3

But it's true that the second version seems more common in our repos, so I'll change it.
Although I would totally postpone all these kind of discussions and changes until we are completely done with the real issues and have our contracts and tests ready, then we agree on a style (I told you I'm willing to accept almost everything) and make all those cosmetic changes, put the lace and send them to auditors.

contracts/court/Court.sol
bingen: Why? Have we agreed on something about it?
I personally don't like implicit returns, but we use them extensively in our codebases and it seems they are slightly more efficient.

contracts/court/Court.sol
facuspagnuolo: I prefer not leaving this in an inconsistent state until we finish with the style discussion, we can change it after if agree on a style that it's not the one we are following currently

test/court/court-appeal.js
bingen: This is used for test titled `'does not modify the current round of the dispute'`.

contracts/court/Court.sol
facuspagnuolo: Again, I wouldn't wait until we agree sth for Aragon in general, I'd love to be consistent in the court codebase and change it afterward if we need to. As you can see, we don't use it in the court unless it is explicitly necessary

test/court/court-initialize.js
bingen: What do you mean properly parsed? This getter is already new. I had to turn `courtConfigs` from `public` to `internal`.

test/court/court-config.js
bingen: I plan to extend this one quite a bit with the second part of configs stuff (issues #36, #53, #90, #127)

test/court/court-initialize.js
facuspagnuolo: Sth like
```js
getCourtConfig(termId) {
   const { pcts, ... } = await court.getCourtConfig(termId)
   return { penaltyPct: pcts[0], ... }
}
```

contracts/court/Court.sol
bingen: Yes, yes, I said I would change it (although I insist, it was consistent before)

contracts/court/Court.sol
bingen: What do you mean "explicitly necessary"? 
Again, we don't use it because [you changed it](https://github.com/aragon/aragon-court/pull/103#discussion_r321631943), and it was used before because we use it extensively in aragonOS, and I think we all agreed that we would try to keep them consistent altogether, not individually.

test/court/court-initialize.js
bingen: Oh, you mean a getter in tests! Ok, I thought you meant a solidity getter. Sure, I'll do that!

contracts/court/Court.sol
bingen: > We could use `uint256` for some of these variables ...

The idea is to save storage slots in structs.

> Additionally `firstRoundJurorsNumber` and `appealStepFactor` could be a uint64 based on how we are using them.

Again, saving storage slots and allowing to be bundled in an array with `maxRegularAppealRounds`. I don't think there's any need for any of them to be greater than `2^16`.

facuspagnuolo: I know we are saving some storage slots, but I think this trade-off is useful when the user is the one paying the stores. Here, I think we are the ones paying for, indeed we are increasing the costs of operating with these values, which is paid by the users.

contracts/court/Court.sol
facuspagnuolo:
```suggestion
    string private constant ERROR_BAD_INITIAL_JURORS = "CT_BAD_INITIAL_JURORS_NUMBER";
```

contracts/court/Court.sol
```diff
@@ -728,6 +741,48 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    *         firstRoundJurorsNumber Number of jurors to be drafted for the first round of disputes
+    *         appealStepFactor Increasing factor for the number of jurors of each round of a dispute
+    *         maxRegularAppealRounds Number of regular appeal rounds before the final round is triggered
+    * @return appealCollateralParams Array containing params for appeal collateral:
+    *         appealCollateralFactor Multiple of juror fees required to appeal a preliminary ruling
+    *         appealConfirmCollateralFactor Multiple of juror fees required to confirm appeal
+    */
+    function getCourtConfig(uint64 _termId) external view
+        returns (
+            address feeToken,
```
facuspagnuolo:
```suggestion
            ERC20 feeToken,
```

contracts/court/Court.sol
```diff
@@ -728,6 +741,48 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+            uint64[4] memory roundStateDurations, //commitTerms, revealTerms, appealTerms, appealConfirmTerms,
+            uint16[2] memory pcts, // penaltyPct, finalRoundReduction,
+            uint64[3] memory roundParams, // firstRoundJurorsNumber, appealStepFactor, maxRegularAppealRounds,
+            uint256[2] memory appealCollateralParams // appealCollateralFactor, appealConfirmCollateralFactor
+        )
+    {
+        CourtConfig storage config = _getConfigAt(_termId);
+        FeesConfig storage feesConfig = config.fees;
+        DisputesConfig storage disputesConfig = config.disputes;
+        feeToken = address(feesConfig.token);
```
facuspagnuolo:
```suggestion
        feeToken = feesConfig.token;
```

contracts/court/Court.sol
```diff
@@ -295,8 +309,8 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         uint256 _minJurorsActiveBalance,
         uint64[4] memory _roundStateDurations,
         uint16[2] memory _pcts, //_penaltyPct, _finalRoundReduction
-        uint64 _appealStepFactor,
-        uint32 _maxRegularAppealRounds,
+        uint64[3] memory _roundParams, // _firstRoundJurorsNumber, _appealStepFactor, _maxRegularAppealRounds
+        uint256[2] memory _appealCollateralParams, // _appealCollateralFactor, _appealConfirmCollateralFactor
```
facuspagnuolo: Nit: I think we should now couple `_maxRegularAppealRounds` to the "appeal" params 

If we change it, I'd update the external getter and internal setter below

contracts/court/Court.sol
```diff
@@ -1157,23 +1217,33 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
-            appealStepFactor: _appealStepFactor,
-            maxRegularAppealRounds: _maxRegularAppealRounds
+            finalRoundReduction: _pcts[1],
+            firstRoundJurorsNumber: _firstRoundJurorsNumber,
+            appealStepFactor: _roundParams[1],
+            maxRegularAppealRounds: _maxRegularAppealRounds,
+            appealCollateralFactor: _appealCollateralParams[0],
+            appealConfirmCollateralFactor: _appealCollateralParams[1]
         });
+        courtConfigs[courtConfigId].disputes = disputesConfig;
```
facuspagnuolo: Don't know if it's less redundant working with storage pointers here, I'm thinking of sth like:

```js
uint64 courtConfigId = uint64(courtConfigs.length++);
CourtConfig storage config = courtConfigs[courtConfigId];

config.fees = FeesConfig({...});
config.disputes = DisputesConfig({...});
```

contracts/court/Court.sol
```diff
@@ -1285,12 +1355,13 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
-        uint64 currentRoundAppealStartTerm = _round.draftTermId + _round.delayedTerms + config.commitTerms + config.revealTerms;
+        uint64 currentRoundAppealStartTerm = _round.draftTermId + _round.delayedTerms +
+            config.disputes.commitTerms + config.disputes.revealTerms;
         // Next round start term is current round end term
-        nextRoundStartTerm = currentRoundAppealStartTerm + config.appealTerms + config.appealConfirmTerms;
+        nextRoundStartTerm = currentRoundAppealStartTerm + config.disputes.appealTerms + config.disputes.appealConfirmTerms;
 
         // Compute next round settings depending on if it will be the final round or not
-        if (_roundId >= uint256(config.maxRegularAppealRounds) - 1) {
+        if (_roundId >= config.disputes.maxRegularAppealRounds - 1) {
```
facuspagnuolo: 🙌

contracts/court/Court.sol
```diff
@@ -1308,8 +1379,18 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         }
 
         // Calculate appeal collateral
-        appealDeposit = totalFees * APPEAL_COLLATERAL_FACTOR;
-        confirmAppealDeposit = totalFees * APPEAL_CONFIRMATION_COLLATERAL_FACTOR;
+        appealDeposit = totalFees * config.disputes.appealCollateralFactor;
+        confirmAppealDeposit = totalFees * config.disputes.appealConfirmCollateralFactor;
```
facuspagnuolo: should we use `SafeMath` here now that we are relying on a variable?

test/court/court-initialize.js
```diff
@@ -26,13 +26,15 @@ contract('Court', () => {
...
+        const termId = await court.getLastEnsuredTermId()
+        const { penaltyPct } = await courtHelper.getCourtConfig(termId)
         assert.equal(penaltyPct.toString(), 0, 'penalty pct does not match')
       })
 
       it('can be 100%', async () => {
         const court = await courtHelper.deploy({ penaltyPct: bn(10000) })
-        const penaltyPct = (await court.courtConfigs(1))[9] // config ID 0 is used for undefined
+        const termId = await court.getLastEnsuredTermId()
+        const { penaltyPct } = await courtHelper.getCourtConfig(termId)
```
facuspagnuolo: 🙌

contracts/court/Court.sol
bingen: Conceptually I like it better as it is, but I prefer having the consistency of `uint256` now that we did this change.

bingen: Hm, the problem is that it's not "appeal" params, it's more "collateral" params. It would feel really weird to move this one but leave `appealStepFactor` with `roundParams`.

PR #150, opened 2019-09-30
Court: expose external config setter

### LINKED ISSUE #36: Court: Add term limits to config changes
- Enforce that config changes need to be scheduled at least a certain number of terms in advance. This limit (`MIN_SCHEDULE_CONFIG_CHANGE_TERMS`) should be a constant in the Court and needs to be higher or equal to the number of terms that a dispute (or appeal round) can be scheduled in advance.

- At the moment when changing the config, if there was already a scheduled config change, it is overridden by the new incoming config. This can be problematic if the term in which the config was scheduled to change is close enough to the current term that there can be disputes or appeal rounds initiated with the fee schedule of a future config which can be cancelled by scheduling a new config change. Config changes shouldn't be possible if there is a pending config change scheduled to happen in less than `MIN_SCHEDULE_CONFIG_CHANGE_TERMS`

### LINKED ISSUE #53: Court: Expose external config setter
Currently it is just an internal function used to set the initial configuration, but the `governor` should be able to change the config. 

There should be a threshold of terms during which no config changes are allowed, if there is an incoming configuration change happening within the threshold, a configuration change overriding it should not be allowed either.

The logic for changing the config after setting it for the first time should also be reviewed, as I believe that both https://github.com/aragon/aragon-court/blob/24cb0916000f9b479842e810485330715c45f671/contracts/Court.sol#L1094 and https://github.com/aragon/aragon-court/blob/24cb0916000f9b479842e810485330715c45f671/contracts/Court.sol#L1119 should use `_fromTermId` instead of `configChangeTermId`

### LINKED ISSUE #90: Court: Improve court settings change flow

### LINKED ISSUE #127: Court: Limit the heartbeat transitions up to the next court config if scheduled

bingen:
Fixes #36
Fixes #53
Fixes #90
Fixes #127

It assumes that config has to remain constant throughout all dispute
lifetime. It seems more fair and the logic is simpler.

TODO:

- [x] Either implement cancel dispute or prevent disputes from being
created in the future (I'd rather do the latter). This is as a
"defensive" mechanism in case the config is changed in between the
dispute is created and it's first draft termId (which is used to get
the config).
- [ ] Fix EIP-170 again.


test/court/court-config.js
```diff
@@ -44,43 +70,245 @@ contract('Court config', () => {
     })
   })
 
+  const checkConfig = async (termId, {
```
izqui: This object here looks a bit ugly.

I propose we just have `const checkConfig = async (termId, newConfig) =>` and then we destructure inside the function body, even in multiple destructurings.

test/court/court-config.js
```diff
@@ -44,43 +70,245 @@ contract('Court config', () => {
...
+
+      const { promise } = await changeConfigPromise(configChangeTermId, courtHelper.governor)
+      await assertRevert(promise, ERROR_TOO_OLD_TERM)
+    })
+
+    it('fails setting config with only one term in advance', async () => {
+      const configChangeTermId = 3
+      // move forward
+      await courtHelper.setTerm(configChangeTermId - 1)
+
```
izqui: pls fix spacing in this entire file 🙏 (ps: we need a linter for test code)

test/court/court-config.js
```diff
@@ -44,43 +70,245 @@ contract('Court config', () => {
...
+    assertBn(penaltyPct, newPenaltyPct, 'Penalty permyriad does not match')
+    assertBn(finalRoundReduction, newFinalRoundReduction, 'Final round reduction does not match')
+    assertBn(firstRoundJurorsNumber, newFirstRoundJurorsNumber, 'First round jurors number does not match')
+    assertBn(appealStepFactor, newAppealStepFactor, 'Appeal step factor does not match')
+    assertBn(maxRegularAppealRounds, newMaxRegularAppealRounds, 'Number af max regular appeal rounds does not match')
+    assertBn(appealCollateralFactor, newAppealCollateralFactor, 'Appeal collateral factor does not match')
+    assertBn(appealConfirmCollateralFactor, newAppealConfirmCollateralFactor, 'Appeal confirmation collateral factor does not match')
+  }
+
+  const buildNewConfig = async (iteration = 1) => {
```
izqui: 🤩 super cool way to test this

test/court/court-config.js
```diff
@@ -44,43 +70,245 @@ contract('Court config', () => {
...
+  }
+
+  const changeConfig = async (termId, iteration = 1) => {
+    const { promise, newConfig } = await changeConfigPromise(termId, courtHelper.governor)
+    await promise
+
+    return newConfig
+  }
+
   context('initialization', () => {
```
izqui: Kind of weird that the tests start here, I think we should move most of the code above to a helper file

contracts/court/Court.sol
```diff
@@ -70,9 +71,6 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     string private constant ERROR_WONT_REWARD_INCOHERENT_JUROR = "CT_WONT_REWARD_INCOHERENT_JUROR";
     string private constant ERROR_ROUND_APPEAL_ALREADY_SETTLED = "CT_APPEAL_ALREADY_SETTLED";
 
-    // Initial term to start the Court, disputes are not allowed during this term. It can be used to active jurors
-    uint64 internal constant ZERO_TERM_ID = 0;
```
izqui: I kinda liked having it explicit instead of just using a `0` as a magic number in the code

contracts/court/Court.sol
```diff
@@ -329,17 +327,63 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+        uint256[4] calldata _fees, // _jurorFee, _heartbeatFee, _draftFee, _settleFee
+        uint64[4] calldata _roundStateDurations,
+        uint16[2] calldata _pcts, //_penaltyPct, _finalRoundReduction
+        uint64[3] calldata _roundParams, // _firstRoundJurorsNumber, _appealStepFactor, _maxRegularAppealRounds
+        uint256[2] calldata _appealCollateralParams // _appealCollateralFactor, _appealConfirmCollateralFactor
+    )
+        external
+        only(governor)
+    {
+        require(_termId > termId, ERROR_PAST_TERM);
```
izqui: As discussed in some of the issues, there should be a number of terms in the future for which config changes are not allowed, so people can see the changes and leave or stop using the Court if they disagree.

contracts/court/Court.sol
```diff
@@ -955,8 +1007,12 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             uint64 currentTermId = termId;
             Term storage currentTerm = terms[currentTermId];
 
-            // TODO: allow config to be changed for a future term id
-            currentTerm.courtConfigId = previousTerm.courtConfigId;
+            // Check if it's time for a config change (only one can be scheduled)
+            if (currentTermId == configChangeTermId) {
+                currentTerm.courtConfigId = previousTerm.courtConfigId + 1;
```
izqui: ⚠️ Won't this break if a config change is scheduled, and it's overwritten by another config scheduled for a prior term?

izqui: We should add an explicit test for this in any case!

contracts/court/Court.sol
```diff
@@ -665,7 +665,12 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         emit AppealDepositSettled(_disputeId, _roundId);
 
         // If the appeal wasn't confirmed, return the entire deposit to appeal maker
-        (,,,ERC20 feeToken, uint256 totalFees,, uint256 appealDeposit, uint256 confirmAppealDeposit) = _getNextRoundDetails(dispute, round, _roundId);
+        (,,,
```
izqui: This doesn't look too good to me 🤷‍♀️

contracts/court/Court.sol
```diff
@@ -239,26 +239,6 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         _;
     }
 
-    /**
```
izqui: I imagine this is a placeholder to make it compile. I think we should revert it once fixed.

contracts/court/Court.sol
```diff
@@ -425,7 +405,8 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
      * @param _disputeId Identification number of the dispute to be drafted
      * @param _maxJurorsToBeDrafted Max number of jurors to be drafted, it will be capped to the requested number of jurors of the dispute
      */
-    function draft(uint256 _disputeId, uint64 _maxJurorsToBeDrafted) external disputeExists(_disputeId) {
+    function draft(uint256 _disputeId, uint64 _maxJurorsToBeDrafted) external {
+        disputeExists(_disputeId);
```
izqui: I prefer the modifier, but in any case I'd rename these to `ensureX` (eg `ensureDisputeExists`)

contracts/court/Court.sol
bingen: What do you mean? As we talked, this is to fall down below EIP-170 limit again. It's in a separate commit so we ca easily revert and choose an alternative if you want.

contracts/court/Court.sol
bingen: To me neither, actually. It has to be broken down to pass linter (line too long). I tried to follow what I saw [here](https://github.com/aragon/aragon-court/blob/development/contracts/court/Court.sol#L474), but let me know how you like it.

contracts/court/Court.sol
bingen: Right now it's just two (this is outdated), so when you create the dispute you can be sure that no unknown settings will be applied to it, and then those settings apply to the whole life cycle of the dispute. Isn't that enough?

contracts/court/Court.sol
bingen: I guess that by magic you mean something like arbitrary. It is not, the way code was, it has to be zero. If you keep the constant `ZERO_TERM_ID` but change its value to, let's say, `1`, it won't work. You should do `termId = ZERO_TERM_ID` first thing in the constructor.
It's quite common to use zero as the first element of arrays (or mappings pretending arrays), so I don't see the need to use a constant for it.
Besides, in `CourtSubcscriptions` we can read:
```
    // Term 0 is for jurors on-boarding
    uint64 internal constant START_TERM_ID = 1;
```
If we keep that constant, CourtSubscriptions should fetch it and use it here.

contracts/court/Court.sol
```diff
@@ -329,50 +305,92 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    */
+    function setCourtConfig(
+        uint64 _termId,
+        ERC20 _feeToken,
+        uint256[4] calldata _fees, // _jurorFee, _heartbeatFee, _draftFee, _settleFee
+        uint64[4] calldata _roundStateDurations,
+        uint16[2] calldata _pcts, //_penaltyPct, _finalRoundReduction
+        uint64[3] calldata _roundParams, // _firstRoundJurorsNumber, _appealStepFactor, _maxRegularAppealRounds
+        uint256[2] calldata _appealCollateralParams // _appealCollateralFactor, _appealConfirmCollateralFactor
+    )
```
facuspagnuolo: Nit: I don't think it is necessary to denote here again which is the stuff being held in each array

contracts/court/Court.sol
```diff
@@ -329,50 +305,92 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    {
+        _setCourtConfig(
+            _termId,
+            _feeToken,
+            _fees,
+            _roundStateDurations,
+            _pcts,
+            _roundParams,
+            _appealCollateralParams
+        );
```
facuspagnuolo: Nit: I think this one fits in a single line, it would be a 111-chars-length sentence if I'm not wrong

contracts/court/Court.sol
```diff
@@ -329,50 +305,92 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    /**
+    * @notice Create a dispute over `_subject` with `_possibleRulings` possible rulings in next term
     * @dev Create a dispute to be drafted in a future term
     * @param _subject Arbitrable subject being disputed
     * @param _possibleRulings Number of possible rulings allowed for the drafted jurors to vote on the dispute
-    * @param _draftTermId Term from which the the jurors for the dispute will be drafted
     * @return Dispute identification number
     */
-    function createDispute(IArbitrable _subject, uint8 _possibleRulings, uint64 _draftTermId) external ensureTerm
+    function createDispute(IArbitrable _subject, uint8 _possibleRulings) external ensureTerm
```
facuspagnuolo: Nit: I think this one now fits in one line as well

contracts/court/Court.sol
```diff
@@ -1367,36 +1422,26 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
 
-    /**
-    * @dev Internal function to calculate the jurors number for the first regular round of a dispute.
-    * @param _draftTermId Term from which the the jurors for the dispute will be drafted
-    * @return Jurors number for the first regular round of a dispute
-    */
-    function _getFirstRoundJurorsNumber(uint64 _draftTermId) internal view returns (uint64) {
-        CourtConfig storage config = _getConfigAt(_draftTermId);
-        return config.disputes.firstRoundJurorsNumber;
-    }
```
facuspagnuolo: Just curious, is there a reason for inlining this code duplicating it?

contracts/court/Court.sol
```diff
@@ -928,11 +959,20 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     * @return weight Juror weight drafted for the requested round
     * @return rewarded Whether or not the given juror was rewarded based on the requested round
     */
-    function getJuror(uint256 _disputeId, uint256 _roundId, address _juror) external view roundExists(_disputeId, _roundId)
+    function getJuror(uint256 _disputeId, uint256 _roundId, address _juror) external view
```
facuspagnuolo: Nit: this one now fits in a single line

contracts/court/Court.sol
```diff
@@ -1260,6 +1313,23 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    }
+
+    /**
+    * @dev Ensure a dispute round exists
+    * @param dispute Dispute the round should belong to
+    * @param _roundId Identification number of the dispute round to be ensured
+    */
+    function roundExists(Dispute storage dispute, uint256 _roundId) internal view {
+        require(_roundId < dispute.rounds.length, ERROR_ROUND_DOES_NOT_EXIST);
+    }
```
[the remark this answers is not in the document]
facuspagnuolo: Another comment in case you are following this strategy... I think you're always checking and then loading from storage, you may want to do both here

contracts/court/Court.sol
```diff
@@ -1177,8 +1226,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     }
 
     /**
-    * TODO: Expose external function to change config
-    * TODO: document
+    * @dev Assumes that sender it's allowed (either it's from governor or it's on init)
```
facuspagnuolo: Could you please complete this with the params doc?

contracts/court/Court.sol
```diff
@@ -1131,22 +1173,30 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
     * @param _round Dispute round to calculate the juror's weight of
     * @param _juror Address of the juror to calculate the weight of
     * @return Weight of the requested juror for the final round of the given dispute
     */
-    function _computeJurorWeightForFinalRound(AdjudicationRound storage _round, address _juror) internal returns (uint64) {
+    function _computeJurorWeightForFinalRound(
+        CourtConfig storage _config,
+        AdjudicationRound storage _round,
+        address _juror
+    )
```
facuspagnuolo: Nit: we can wrap the params and the modifiers in the same line as we are doing in the rest of the functions that are being split into multiple lines

contracts/court/Court.sol
```diff
@@ -1131,22 +1173,30 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
             ? _getJurorWeightForRegularRound(round, _juror)
-            : _computeJurorWeightForFinalRound(round, _juror);
+            : _computeJurorWeightForFinalRound(config, round, _juror);
     }
 
     /**
     * @dev Internal function to compute the juror weight for the final round. Note that for a final round the weight of
     *      each juror is equal to the number of times the min active balance the juror has. This function will try to
     *      collect said amount from the active balance of a juror, acting as a lock to allow them to vote.
+    * @param _config Court config to use in order to compute weight
```
facuspagnuolo:
```suggestion
    * @param _config Court config to calculate the juror's weight of
```

test/court/court-config.js
```diff
@@ -1,15 +1,22 @@
...
+  let originalConfig, initialConfig
   let feeToken
   let jurorFee, heartbeatFee, draftFee, settleFee
   let commitTerms, revealTerms, appealTerms, appealConfirmTerms
   let penaltyPct, finalRoundReduction
   let firstRoundJurorsNumber, appealStepFactor, maxRegularAppealRounds
   let appealCollateralFactor, appealConfirmCollateralFactor
 
+  const ERROR_BAD_SENDER = 'CT_BAD_SENDER'
+  const ERROR_TOO_OLD_TERM = 'CT_TOO_OLD_TERM'
```
facuspagnuolo: I like this, we can have a central errors object like we are doing in the apps

contracts/court/Court.sol
```diff
@@ -1213,12 +1263,15 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             require(_roundStateDurations[i] > 0, ERROR_CONFIG_PERIOD_ZERO_TERMS);
         }
 
-        if (configChangeTermId != ZERO_TERM_ID) {
-            terms[configChangeTermId].courtConfigId = 0; // reset previously set fee structure change
+        // reset previously set fee structure future change
+        // in that case we will overwrite last array item
+        if (configChangeTermId > termId) {
+            terms[configChangeTermId].courtConfigId = 0;
+        } else { // otherwise increase configs array for the new one
```
facuspagnuolo:
```suggestion
        } else { 
```

contracts/court/Court.sol
```diff
@@ -1213,12 +1263,15 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             require(_roundStateDurations[i] > 0, ERROR_CONFIG_PERIOD_ZERO_TERMS);
         }
 
-        if (configChangeTermId != ZERO_TERM_ID) {
-            terms[configChangeTermId].courtConfigId = 0; // reset previously set fee structure change
+        // reset previously set fee structure future change
```
facuspagnuolo:
```suggestion
        // If there was a config change already scheduled, reset it. Otherwise, schedule a new config.
```

contracts/court/Court.sol
```diff
@@ -1213,12 +1263,15 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+        // in that case we will overwrite last array item
```
facuspagnuolo:
```suggestion
```

contracts/court/Court.sol
```diff
@@ -1213,12 +1263,15 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+        if (configChangeTermId > termId) {
+            terms[configChangeTermId].courtConfigId = 0;
```
facuspagnuolo: We could cache `configChangeTermId `

test/helpers/court.js
```diff
@@ -220,6 +220,114 @@ module.exports = (web3, artifacts) => {
...
+      assertBn(appealTerms, newAppealTerms, 'Appeal terms number does not match')
+      assertBn(appealConfirmTerms, newAppealConfirmTerms, 'Appeal confirmation terms number does not match')
+      assertBn(penaltyPct, newPenaltyPct, 'Penalty permyriad does not match')
+      assertBn(finalRoundReduction, newFinalRoundReduction, 'Final round reduction does not match')
+      assertBn(firstRoundJurorsNumber, newFirstRoundJurorsNumber, 'First round jurors number does not match')
+      assertBn(appealStepFactor, newAppealStepFactor, 'Appeal step factor does not match')
+      assertBn(maxRegularAppealRounds, newMaxRegularAppealRounds, 'Number af max regular appeal rounds does not match')
+      assertBn(appealCollateralFactor, newAppealCollateralFactor, 'Appeal collateral factor does not match')
+      assertBn(appealConfirmCollateralFactor, newAppealConfirmCollateralFactor, 'Appeal confirmation collateral factor does not match')
+    }
```
facuspagnuolo: I don't think the assertion stuff belongs here, we can either have a new helper or leave it in the file where its being used if it is only used once

test/court/court-draft.js
```diff
@@ -342,7 +343,9 @@ contract('Court', ([_, disputer, drafter, juror500, juror1000, juror1500, juror2
           context('when the court term is outdated by one term', () => {
             beforeEach('ensure previous term of the draft term', async () => {
               const neededTransitions = await court.neededTermTransitions()
-              await court.heartbeat(neededTransitions.sub(bn(1)))
+              if (neededTransitions.toNumber() > 1) {
+                await court.heartbeat(neededTransitions.sub(bn(1)))
+              }
```
facuspagnuolo: Why was this necessary?

test/court/court-disputes.js
```diff
@@ -72,12 +76,14 @@ contract('Court', ([_, sender]) => {
           })
 
           it('transfers fees to the court', async () => {
+            // move forward to the term before the desired start one for the dispute
+            await courtHelper.setTerm(draftTermId - 1)
```
facuspagnuolo: Why do we need this for the first three tests if we are already doing it in the `beforeEach` block below?

test/court/court-disputes.js
```diff
@@ -189,11 +180,12 @@ contract('Court', ([_, sender]) => {
       const possibleRulings = 2
 
       beforeEach('create dispute', async () => {
-        await courtHelper.setTimestamp(firstTermStartTime)
+        // move forward to the term before the desired start one for the dispute
+        await courtHelper.setTerm(draftTermId - 1)
```
facuspagnuolo: I think it is better to mock the current term directly instead of doing it based on a draft term since we are not relying on it any more

test/court/court-settle.js
```diff
@@ -37,9 +37,8 @@ contract('Court', ([_, disputer, drafter, appealMaker, appealTaker, juror500, ju
       beforeEach('activate jurors and create dispute', async () => {
         await courtHelper.activate(jurors)
 
-        await courtHelper.setTerm(1)
         disputeId = await courtHelper.dispute({ draftTermId, disputer })
-        await courtHelper.passTerms(bn(draftTermId - 1)) // court is already at term one
+        await courtHelper.passTerms(bn(1)) // court is already at term previous to dispute start
```
facuspagnuolo:
```suggestion
        await courtHelper.passTerms(bn(1)) // pass one term to reach the draft term
```

test/court/court-voting.js
```diff
@@ -33,9 +33,8 @@ contract('Court', ([_, disputer, drafter, juror100, juror500, juror1000, juror15
     beforeEach('activate jurors and create dispute', async () => {
       await courtHelper.activate(jurors)
 
-      await courtHelper.setTerm(1)
       disputeId = await courtHelper.dispute({ draftTermId, disputer })
-      await courtHelper.passTerms(bn(draftTermId - 1)) // court is already at term one
+      await courtHelper.passTerms(bn(1)) // court is already at term previous to dispute start
```
facuspagnuolo:
```suggestion
      await courtHelper.passTerms(bn(1)) // pass one term to reach the draft term
```

test/court/court-config.js
```diff
@@ -46,41 +81,176 @@ contract('Court config', () => {
...
+
+      const { promise } = await courtHelper.changeConfigPromise(originalConfig, configChangeTermId, courtHelper.governor)
+      await assertRevert(promise, ERROR_TOO_OLD_TERM)
+    })
+
+    context('governor can change config', () => {
+      const configChangeTermId = 3
+      let newConfig
+
+      beforeEach('ask for the change', async () => {
```
facuspagnuolo:
```suggestion
      beforeEach('change court config', async () => {
```

test/court/court-config.js
```diff
@@ -46,41 +81,176 @@ contract('Court config', () => {
...
+    it('fails setting config with only one term in advance', async () => {
+      const configChangeTermId = 3
+      // move forward
+      await courtHelper.setTerm(configChangeTermId - 1)
+
+      const { promise } = await courtHelper.changeConfigPromise(originalConfig, configChangeTermId, courtHelper.governor)
+      await assertRevert(promise, ERROR_TOO_OLD_TERM)
+    })
+
+    context('governor can change config', () => {
```
facuspagnuolo: I'd wrap the tests above in a `context('when the config change fails')` block and wrap the following tests into

```suggestion
    context('when the config change succeeds', () => {
```

test/court/court-config.js
```diff
@@ -46,41 +81,176 @@ contract('Court config', () => {
...
+      it('check once the change term id has been reached', async () => {
+        // move forward
+        await courtHelper.setTerm(configChangeTermId)
+
+        await courtHelper.checkConfig(configChangeTermId, newConfig)
+      })
+    })
+
+    context('overwriting changes at a later term', () => {
+      const configChangeTermId1 = 3, configChangeTermId2 = 4
```
facuspagnuolo: What about simply adding a `newChangeTermId` variable here instead? It could be placed within the context above

test/helpers/court.js
```diff
@@ -220,6 +220,114 @@ module.exports = (web3, artifacts) => {
...
+        newJurorFee, newHeartbeatFee, newDraftFee, newSettleFee,
+        newCommitTerms, newRevealTerms, newAppealTerms, newAppealConfirmTerms,
+        newPenaltyPct, newFinalRoundReduction,
+        newFirstRoundJurorsNumber, newAppealStepFactor, newMaxRegularAppealRounds,
+        newAppealCollateralFactor,
+        newAppealConfirmCollateralFactor,
+      }
+    }
+
+    async changeConfigPromise(originalConfig, termId, from, iteration = 1) {
```
facuspagnuolo: I think `iteration` is never used

test/helpers/court.js
```diff
@@ -220,6 +220,114 @@ module.exports = (web3, artifacts) => {
...
+        [ newPenaltyPct, newFinalRoundReduction ],
+        [ newFirstRoundJurorsNumber, newAppealStepFactor, newMaxRegularAppealRounds ],
+        [ newAppealCollateralFactor, newAppealConfirmCollateralFactor ],
+        { from }
+      )
+
+      return { promise, newConfig }
+    }
+
+    async changeConfig(originalConfig, termId, iteration = 1) {
```
facuspagnuolo: I think `iteration` is never used here as well

test/helpers/court.js
```diff
@@ -220,6 +220,114 @@ module.exports = (web3, artifacts) => {
...
+      const { promise, newConfig } = await this.changeConfigPromise(originalConfig, termId, this.governor)
+      await promise
+
+      return newConfig
```
facuspagnuolo: Why not simply awaiting the change always? You won't get a `newConfig` if it fails, and you could test the revert doing simply `await assertRevert(courtHelper.changeConfig(...), 'error')`

test/helpers/court.js
```diff
@@ -220,6 +220,114 @@ module.exports = (web3, artifacts) => {
...
+      assertBn(penaltyPct, newPenaltyPct, 'Penalty permyriad does not match')
+      assertBn(finalRoundReduction, newFinalRoundReduction, 'Final round reduction does not match')
+      assertBn(firstRoundJurorsNumber, newFirstRoundJurorsNumber, 'First round jurors number does not match')
+      assertBn(appealStepFactor, newAppealStepFactor, 'Appeal step factor does not match')
+      assertBn(maxRegularAppealRounds, newMaxRegularAppealRounds, 'Number af max regular appeal rounds does not match')
+      assertBn(appealCollateralFactor, newAppealCollateralFactor, 'Appeal collateral factor does not match')
+      assertBn(appealConfirmCollateralFactor, newAppealConfirmCollateralFactor, 'Appeal confirmation collateral factor does not match')
+    }
+
+    async buildNewConfig (originalConfig, iteration = 1) {
```
facuspagnuolo: I think `iteration` is always 1 here

test/court/court-config.js
```diff
@@ -46,41 +81,176 @@ contract('Court config', () => {
...
+      { address: juror3000, initialActiveBalance: bigExp(3000, 18) },
+      { address: juror500,  initialActiveBalance: bigExp( 500, 18) },
+      { address: juror1000, initialActiveBalance: bigExp(1000, 18) },
+    ]
+
+    beforeEach('activate jurors', async () => {
+      await courtHelper.activate(jurors)
+    })
+
+    it('dispute is not affected by config settings during its lifetime', async () => {
```
facuspagnuolo:
```suggestion
    it('does not affect a dispute during its lifetime', async () => {
```

contracts/court/Court.sol
```diff
@@ -329,50 +305,92 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    *        _finalRoundReduction ‱ of fee reduction for the last appeal round (1/10,000)
+    * @param _roundParams Array containing params for rounds:
+    *        _firstRoundJurorsNumber Number of jurors to be drafted for the first round of disputes
+    *        _appealStepFactor Increasing factor for the number of jurors of each round of a dispute
+    *        _maxRegularAppealRounds Number of regular appeal rounds before the final round is triggered
+    * @param _appealCollateralParams Array containing params for appeal collateral:
+    *        _appealCollateralFactor Multiple of juror fees required to appeal a preliminary ruling
+    *        _appealConfirmCollateralFactor Multiple of juror fees required to confirm appeal
+    */
+    function setCourtConfig(
```
facuspagnuolo: Nit: I'd place this function below the rest of the external core Court functions like dispute, appeal, confirm appeal, etc

contracts/court/Court.sol
```diff
@@ -329,50 +305,92 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
     * @param _subject Arbitrable subject being disputed
     * @param _possibleRulings Number of possible rulings allowed for the drafted jurors to vote on the dispute
-    * @param _draftTermId Term from which the the jurors for the dispute will be drafted
     * @return Dispute identification number
     */
-    function createDispute(IArbitrable _subject, uint8 _possibleRulings, uint64 _draftTermId) external ensureTerm
+    function createDispute(IArbitrable _subject, uint8 _possibleRulings) external ensureTerm
         returns (uint256)
     {
         // TODO: Limit the min amount of terms before drafting (to allow for evidence submission)
```
facuspagnuolo: Is this one still necessary ?

contracts/court/Court.sol
facuspagnuolo: We can do sth like [this](https://github.com/aragon/aragon-court/pull/155)

facuspagnuolo:
For the record, feel free to ignore all the comments you think are related to styling. We can tackle those once we have finished with all the issues

contracts/court/Court.sol
bingen: I have said several times (offline and in my previous comment) that this was temporary, only meant to pass CI.  I put it in a separate commit to make it easy to do a `git revert`, so once we fix contract size issue any other way I'll undo it.

contracts/court/Court.sol
bingen: As I said this is going to be reverted.

contracts/court/Court.sol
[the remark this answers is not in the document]
bingen: No, I'm not following this strategy. This is temporary and it's going to be reverted.

contracts/court/Court.sol
bingen: Yes, to avoid fetching `config` twice.

contracts/court/Court.sol
bingen: This is part of the modifiers to-be-reverted commit.

contracts/court/Court.sol
bingen: I was following this:
https://solidity.readthedocs.io/en/v0.5.8/style-guide.html#function-declaration
as I have always done (and as I was doing initially in this repo).
But no problem, reverting.

contracts/court/Court.sol
bingen: If I do it, I hit stack too deep issue. It's a good suggestion though, I did it for `_getConfigAt`.

test/court/court-draft.js
bingen: Without it one tests fails with `CT_INVALID_TRANSITION_TERMS`, as if `neededTransitions` is `1`, we would be trying to `heartbeat` zero transitions.

test/court/court-disputes.js
bingen: I guess because it's below. Do you mean the one in line 184? If I'm not wrong, that block code doesn't fall below that `beforeEach` clause.
Anyway, if you pt this line right before this `setTerm`:
```
            console.log((await court.getLastEnsuredTermId()).toString())
```
You'll see that `termId` is not always `draftTermId - 1`.
Also, if you try to change the previous:
```
      const draftTermId = 2
```
by, let's say, `3`, you'll realize that without that `setTerm` it doesn't work.

test/court/court-disputes.js
bingen: No, because now are always scheduled in the very next term (so they can't be scheduled in advance), so we need to make sure that we are in the previous term to desired `draftTermId`.

test/court/court-settle.js
bingen: `dispute` helper from `court.js` now sets the term to the previous one to the desired dispute draftTermId

test/court/court-voting.js
bingen: `dispute` helper from `court.js` now sets the term to the previous one to the desired dispute draftTermId

test/court/court-config.js
bingen: Sorry, I don't get what you mean.

test/helpers/court.js
bingen: `changeConfigPromise` can now pass it, and some the tests with 2 configs have iterations with value 2.

contracts/court/Court.sol
bingen: Why? To me is closer "semantically" to the constuctor, that's why I put it there.

contracts/court/Court.sol
bingen: Not sure. I wanted to review the agreements part before removing this.

contracts/court/Court.sol
facuspagnuolo: Sure, me too! both styles are fine and both were being used in this repo even before I joined if that's what you mean. My two cents here is pushing for *consistency* 😉 

image

PR #153, opened 2019-10-04
Sortition: fix batch intervals

bingen:
There is no need to add and subtract 1 to lower bound, reverting to
original logic.


contracts/lib/JurorsTreeSortition.sol
```diff
@@ -103,62 +106,30 @@ library JurorsTreeSortition {
         pure
         returns (uint256[] memory)
     {
-        // Calculate the interval to be used to search the balances in the tree. Since we are using a modulo function
-        // to compute the random balances to be searched, we add one to the difference to make sure the last number
-        // of the range is also included. For example, to compute a range [1,10] we need to compute using modulo 10.
```
facuspagnuolo: I wouldn't drop this explanation to avoid a confusion again, I'd just update it using a `[0,10)` range

PR #159, opened 2019-10-09
Chore: Gas reporter

bingen:


.travis.yml
```diff
@@ -37,3 +37,6 @@ jobs:
     - stage: tests
       script: npm test test/subscriptions/*.js
       name: "Subscriptions"
+    - stage: tests
```
facuspagnuolo: WDYT about adding it in a separate stage? If tests are failing it may not make sense to measure gas costs

bingen:
You are right bout the duration. I'm thinking I'm going to remove the CI part, but leave the tool so we can at least run it manually.

bingen:
Closed in favor of #161.

PR #160, opened 2019-10-09
Registry: Prevent sortition interval narrowing

### LINKED ISSUE #156: Registry: Draft iterations shouldn't modify `selectedJurors` and `batchRequestedJurors`
Otherwise we are altering the chances of being selected, being the jurors closer to the bath inferior limit more likely to be selected.
The reason is that those values are used to calculate the bonds of the search.

bingen:
Before `selectedJurors` and `batchRequestedJurors` were modified on
different sortition iterations, resulting in a narrowing of the search
interval, which alters the probability of jurors being selected. Now
those values remain constant through different iterations.

Fixes #156.

### DIFF
--- a/test/registry/jurors-registry-drafting.js
+++ b/test/registry/jurors-registry-drafting.js
@@ -275,11 +275,23 @@ contract('JurorsRegistry', ([_, juror500, juror1000, juror1500, juror2000, juror
             })
 
             assert.equal(outputLength.toNumber(), expectedJurors.length, 'output length does not match')
-            for (let i = 0; i < outputLength.toNumber(); i++) {
-              const currentLockedBalance = (await registry.balanceOf(expectedJurors[i].address))[2]
-              const previousLockedBalance = previousLockedBalances[expectedJurors[i].address]
-              const expectedLockedBalance = expectedJurors[i].weight * (MIN_ACTIVE_AMOUNT.mul(DRAFT_LOCK_PCT).div(bn(10000)))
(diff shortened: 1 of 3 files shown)


test/registry/jurors-registry-drafting.js
```diff
@@ -275,11 +275,23 @@ contract('JurorsRegistry', ([_, juror500, juror1000, juror1500, juror2000, juror
             })
 
             assert.equal(outputLength.toNumber(), expectedJurors.length, 'output length does not match')
-            for (let i = 0; i < outputLength.toNumber(); i++) {
-              const currentLockedBalance = (await registry.balanceOf(expectedJurors[i].address))[2]
-              const previousLockedBalance = previousLockedBalances[expectedJurors[i].address]
-              const expectedLockedBalance = expectedJurors[i].weight * (MIN_ACTIVE_AMOUNT.mul(DRAFT_LOCK_PCT).div(bn(10000)))
-              assert.equal(currentLockedBalance.sub(previousLockedBalance).toString(), expectedLockedBalance, `locked balance for juror #${i} does not match`)
+            // merge possible duplicates
+            const mergedExpectedJurors = expectedJurors.reduce((acc, juror) => {
```
facuspagnuolo: WDTY renaming `acc` to `jurors` or `jurorsInfo`

bingen: Well, with `reduce` I sometimes forget the order of params, so I like to use `acc` to emphasize that it's the accumulator. You'll see it already appears like that in several other tests.  Besides I would say it's a quite common practice.

PR #161, opened 2019-10-10
Add gas reporter

bingen:
Replaces #159.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "compile": "truffle compile",
     "lint": "solium --dir ./contracts",
     "test": "npm run ganache-cli:test",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
     "compile": "truffle compile",
     "lint": "solium --dir ./contracts",
     "test": "npm run ganache-cli:test",
+    "test:gas": "GAS_REPORTER=true npm test",
```
facuspagnuolo: the env variable is required by the tool?

bingen: Nope, it's used by @aragon/os truffle-config.js

PR #162, opened 2019-10-10
Contracts: Add SafeMath

### LINKED ISSUE #22: Contracts: Use SafeMath for balance updates
For instance: 
https://github.com/aragonone/aragon-court/blob/778ab8b28f93911940152f4925acd57f3ad2f4a1/contracts/Court.sol#L969
https://github.com/aragonone/aragon-court/blob/778ab8b28f93911940152f4925acd57f3ad2f4a1/contracts/Court.sol#L1039
https://github.com/aragonone/aragon-court/blob/778ab8b28f93911940152f4925acd57f3ad2f4a1/contracts/Court.sol#L1045
https://github.com/aragonone/aragon-court/blob/778ab8b28f93911940152f4925acd57f3ad2f4a1/contracts/Court.sol#L1051

Or if we are completely sure that those can not over/underflow due to the inner logic, add a comment explaining it.

bingen:
Also added comments to every place where it's not needed explaining
the reason.

Fixes #22.

TODO:

- [ ] Fix EIP-170 once again.

### DIFF
--- a/contracts/court/Court.sol
+++ b/contracts/court/Court.sol
@@ -315,7 +315,10 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     ) public {
         // This seems reasonable enough, and this way we avoid using SafeMath for termDurarion
         require(_termDuration < ONE_YEAR, ERROR_TERM_DURATION_TOO_LONG);
-        require(_firstTermStartTime >= getTimestamp64() + _termDuration, ERROR_BAD_FIRST_TERM_START_TIME);
+        require(
+            _firstTermStartTime >= getTimestamp64() + _termDuration && _firstTermStartTime <= getTimestamp64() + 2 * ONE_YEAR,
+            ERROR_BAD_FIRST_TERM_START_TIME
+        );
 
         termDuration = _termDuration;
         jurorsRegistry = _jurorsRegistry;
@@ -642,8 +645,9 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
 
         // Distribute the collected tokens of the jurors that were slashed weighted by the winning jurors. Note that
         // we are penalizing jurors that refused intentionally their vote for the final round.
+        // Note that coherentJurors has to be > 0 because above it's required that _juror voted in favour (so at the very least there is one),
+        // therefore divisions below are safe.
         uint256 coherentJurors = round.coherentJurors;
-        require(coherentJurors > 0, ERROR_NO_COHERENT_JURORS);
         uint256 collectedTokens = round.collectedTokens;
         if (collectedTokens > 0) {
             jurorsRegistry.assignTokens(_juror, uint256(jurorState.weight).mul(collectedTokens) / coherentJurors);
@@ -693,13 +697,15 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         // Note that we are safe to access the dispute final ruling, since we already ensured that round penalties were settled.
         uint8 finalRuling = dispute.finalRuling;
         uint256 totalDeposit = appealDeposit.add(confirmAppealDeposit);
-        // No need for SafeMath in these subtractions: as collateral factors > 0 => deposits > totalFees
         if (appeal.appealedRuling == finalRuling) {
+            // No need for SafeMath: collateral factors > 0 => both appeal deposits > totalFees
             accounting.assign(feeToken, appeal.maker, totalDeposit - totalFees);
         } else if (appeal.opposedRuling == finalRuling) {
+            // No need for SafeMath: collateral factors > 0 => both appeal deposits > totalFees
             accounting.assign(feeToken, appeal.taker, totalDeposit - totalFees);
         } else {
             uint256 feesRefund = totalFees / 2;
+            // No need for SafeMath: collateral factors > 0 => both appeal deposits > totalFees
             accounting.assign(feeToken, appeal.maker, appealDeposit - feesRefund);
             accounting.assign(feeToken, appeal.taker, confirmAppealDeposit - feesRefund);
         }
@@ -1042,7 +1048,8 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             }
             // Set the start time of the new term. Note that we are using a constant term duration value to guarantee
             // equally long terms, regardless of heartbeats.
-            // No need for SafeMath: termDuration is capped at ONE_YEAR
+            // No need for SafeMath: termDuration is capped at ONE_YEAR, _firstTermStartTime by 2 * ONE_YEAR,
+            // and we assume that timestamps (and its derivatives like termId) won't reach MAX_UINT64, which would be ~5.8e11 years
             currentTerm.startTime = previousTerm.startTime + termDuration;
             // In order to draft a random number of jurors in a term, we use a randomness factor for each term based on a
             // block number that is set once the term has started. Note that this information could not be known beforehand.
@@ -1186,8 +1193,8 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             _round.settledPenalties = true;
         }
 
-        // Update the number of round settled jurors. Note that we don't need to use SafeMath here since the highest number of jurors to be
-        // settled for a round could be the `jurorsNumber` itself, which is a uint64 value.
+        // Update the number of round settled jurors.
+        // No need for SafeMath: the highest number of jurors to be settled for a round could be the `jurorsNumber` itself, which is a uint64.
         _round.settledJurors = uint64(roundSettledJurors + batchSettledJurors);
 
         // Prepare the list of jurors and penalties to either be slashed or returned based on their votes for the given round
@@ -1400,7 +1407,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             return uint64(0);
         }
 
-        // We already know that the start time of the current term is in the past, we are safe to avoid SafeMath here
+        // No need for SafeMath: we already know that the start time of the current term is in the past
         return (getTimestamp64() - currentTermStartTime) / termDuration;
     }
 
@@ -1454,6 +1461,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         DisputesConfig storage disputesConfig = config.disputes;
 
         // No need for SafeMath: round state durations are safely capped at config
+        // and we assume that timestamps (and its derivatives like termId) won't reach MAX_UINT64, which would be ~5.8e11 years
         uint64 currentRoundAppealStartTerm = _round.draftTermId + _round.delayedTerms + disputesConfig.commitTerms + disputesConfig.revealTerms;
         // Next round start term is current round end term
         nextRound.nextRoundStartTerm = currentRoundAppealStartTerm + disputesConfig.appealTerms + disputesConfig.appealConfirmTerms;
@@ -1783,7 +1791,8 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         ];
         (address[] memory jurors, uint64[] memory weights, uint256 outputLength) = jurorsRegistry.draft(draftParams);
 
-        // Update round with drafted jurors information. We are safe to avoid SafeMath here since this cannot be greater than `jurorsNumber`.
+        // Update round with drafted jurors information.
+        // No need for SafeMath: this cannot be greater than `jurorsNumber`.
         uint64 newSelectedJurors = _selectedJurors + _requestedJurors;
         _round.selectedJurors = newSelectedJurors;
 
--- a/contracts/court/Court.sol
+++ b/contracts/court/Court.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.5.8;
 import "@aragon/os/contracts/lib/token/ERC20.sol";
 import "@aragon/os/contracts/common/SafeERC20.sol";
 import "@aragon/os/contracts/lib/math/SafeMath.sol";
+import "@aragon/os/contracts/lib/math/SafeMath64.sol";
 import "@aragon/os/contracts/common/Uint256Helpers.sol";
 import "@aragon/os/contracts/common/TimeHelpers.sol";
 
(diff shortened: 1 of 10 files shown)


contracts/court/Court.sol
```diff
@@ -1431,12 +1453,13 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         CourtConfig storage config = _getDisputeConfig(_dispute);
         DisputesConfig storage disputesConfig = config.disputes;
 
-        // Court terms are assumed to always fit in uint64. Thus, the end term of a round is assumed to fit in uint64 too.
+        // No need for SafeMath: round state durations are safely capped at config
```
facuspagnuolo: Same here, even though they are capped, it can be possible to overflow. IMO we are simply relying on that the term id will never reach max uint64

contracts/court/Court.sol
```diff
@@ -1653,7 +1682,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     * @return Identification number of the Court current term
     */
     function _getCurrentTermId() internal view returns (uint64) {
-        // Court terms are assumed to always fit in uint64. Thus, some terms after the last ensured term is assumed to fit in uint64 too.
```
facuspagnuolo: may sound redundant but I wouldn't remove it

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -565,7 +567,7 @@ contract CourtSubscriptions is TimeHelpers, Initializable, IsContract, ISubscrip
     */
     function _getPeriodStartTermId(uint256 _periodId) internal view returns (uint64) {
         // Periods are measured in Court terms. Since Court terms are represented in `uint64`, we are safe to use `uint64` for period ids too.
-        return START_TERM_ID + uint64(_periodId) * periodDuration;
+        return START_TERM_ID.add(uint64(_periodId).mul(periodDuration));
```
facuspagnuolo: I think this will always fit in `uint64` since we are already assuming that for terms

contracts/court/Court.sol
```diff
@@ -627,13 +643,14 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         // Distribute the collected tokens of the jurors that were slashed weighted by the winning jurors. Note that
         // we are penalizing jurors that refused intentionally their vote for the final round.
         uint256 coherentJurors = round.coherentJurors;
+        require(coherentJurors > 0, ERROR_NO_COHERENT_JURORS);
```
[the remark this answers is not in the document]
bingen: Absolutely, I'll remove it then.

contracts/court/Court.sol
```diff
@@ -675,7 +692,8 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         // selected by any of the appealing parties or no juror voted in the in favor of the possible outcomes, we split it between both parties.
         // Note that we are safe to access the dispute final ruling, since we already ensured that round penalties were settled.
         uint8 finalRuling = dispute.finalRuling;
-        uint256 totalDeposit = appealDeposit + confirmAppealDeposit;
+        uint256 totalDeposit = appealDeposit.add(confirmAppealDeposit);
+        // No need for SafeMath in these subtractions: as collateral factors > 0 => deposits > totalFees
```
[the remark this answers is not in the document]
bingen: The three subtractions inside the if-else if-else clause. But I'll put the comment inside if you think it's clearer this way.

contracts/court/Court.sol
```diff
@@ -1024,6 +1042,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
             }
             // Set the start time of the new term. Note that we are using a constant term duration value to guarantee
             // equally long terms, regardless of heartbeats.
+            // No need for SafeMath: termDuration is capped at ONE_YEAR
```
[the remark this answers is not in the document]
bingen: Well, of course, I was giving that for granted (that timestamp won't reach 5.8e11 years), and what seemed problematic to me was the part coming from external human input (`termDuration`) but I'll clarify this too.

contracts/court/Court.sol
bingen: Same comment as above. 
But we are not *simply* relying on termId not reaching max uint64: if a config with a term dration of ~MAX_UINT64 - 1.6e9 is set, this would end up overflowing.

contracts/court/Court.sol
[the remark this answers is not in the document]
bingen: Actually, in this case term `startTime` is not affected by timestamp, as it comes from adding `termDuration` to `_firstTermStartTime`.

contracts/lib/HexSumTree.sol
```diff
@@ -388,10 +403,17 @@ library HexSumTree {
                 if (_params.level == ITEMS_LEVEL) {
                     _copyFoundNode(_params.foundValues, subtreeIncludedValues, childNodeKey, _resultKeys, childNodeValue, _resultValues);
                 } else {
-                    SearchParams memory nextLevelParams = SearchParams(_params.time, _params.level - 1, childNodeKey, _params.foundValues, _params.visitedTotal);
+                    SearchParams memory nextLevelParams = SearchParams(
+                        _params.time,
+                        _params.level - 1, // No need for SafeMath: ITEMS_LEVEL = 0 => _params.level > 0 here
```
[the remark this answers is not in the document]
bingen: Well, although I personally like it, it's clear and concise IMO, and kind of resembles a formal verification which is what we are trying to do here, here I did it to avoid having a line too long.

contracts/subscriptions/CourtSubscriptions.sol
bingen: If any user calls `getPeriodBalanceDetails` with a too high `_periodId` param it would overflow and therefore return wrong information.

contracts/court/Court.sol
bingen: Yes, indeed it sounds too redundant to me. The result of this operation is itself a court term, right?

contracts/court/Court.sol
```diff
@@ -67,6 +72,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     string private constant ERROR_JUROR_ALREADY_REWARDED = "CT_JUROR_ALREADY_REWARDED";
     string private constant ERROR_WONT_REWARD_NON_VOTER_JUROR = "CT_WONT_REWARD_NON_VOTER_JUROR";
     string private constant ERROR_WONT_REWARD_INCOHERENT_JUROR = "CT_WONT_REWARD_INCOHERENT_JUROR";
+    string private constant ERROR_NO_COHERENT_JURORS = "CT_NO_COHERENT_JURORS";
```
facuspagnuolo: We can drop this one now I think

contracts/court/Court.sol
```diff
@@ -84,6 +90,12 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     // Precision factor used to improve rounding when computing weights for the final round
     uint256 internal constant FINAL_ROUND_WEIGHT_PRECISION = 1000;
 
+    // Max for termDuration (in seconds)
+    uint64 internal constant ONE_YEAR = 31536000;
+
+    // Max number of terms for rond state durations (if terms were 1h, this would be a year)
```
izqui:
```suggestion
    // Max number of terms that each of the different adjudication states can last (if lasted 1h, this would be a year)
```

contracts/court/Court.sol
```diff
@@ -84,6 +90,12 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     // Precision factor used to improve rounding when computing weights for the final round
     uint256 internal constant FINAL_ROUND_WEIGHT_PRECISION = 1000;
 
+    // Max for termDuration (in seconds)
+    uint64 internal constant ONE_YEAR = 31536000;
```
izqui: Nit: let's make it consistent with the `MAX_ADJ_STATE_DURATION` constant
```suggestion
    uint64 internal constant MAX_TERM_DURATION = 1 years;
```

contracts/court/Court.sol
```diff
@@ -84,6 +90,12 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+
+    // Max number of terms for rond state durations (if terms were 1h, this would be a year)
+    uint64 internal constant MAX_STATE_DURATION = 8670;
```
izqui: Nit: use more precise names, 'state' is a super vague word :)
```suggestion
    uint64 internal constant MAX_ADJ_STATE_DURATION = 8670;
```

contracts/court/Court.sol
```diff
@@ -301,7 +313,12 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         uint256[2] memory _appealCollateralParams, // _appealCollateralFactor, _appealConfirmCollateralFactor
         uint256[6] memory _subscriptionParams // _periodDuration, _feeAmount, _prePaymentPeriods, _resumePrePaidPeriods, _latePaymentPenaltyPct, _governorSharePct
     ) public {
-        require(_firstTermStartTime >= getTimestamp64() + _termDuration, ERROR_BAD_FIRST_TERM_START_TIME);
+        // This seems reasonable enough, and this way we avoid using SafeMath for termDurarion
+        require(_termDuration < ONE_YEAR, ERROR_TERM_DURATION_TOO_LONG);
+        require(
+            _firstTermStartTime >= getTimestamp64() + _termDuration && _firstTermStartTime <= getTimestamp64() + 2 * ONE_YEAR,
```
izqui: We should definitely document somewhere that the first term needs to start within two years from contract deployment.

I'd also move this to a constant or at least a named variable as it is a magic number right now

contracts/court/Court.sol
```diff
@@ -675,13 +696,16 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         // selected by any of the appealing parties or no juror voted in the in favor of the possible outcomes, we split it between both parties.
         // Note that we are safe to access the dispute final ruling, since we already ensured that round penalties were settled.
         uint8 finalRuling = dispute.finalRuling;
-        uint256 totalDeposit = appealDeposit + confirmAppealDeposit;
+        uint256 totalDeposit = appealDeposit.add(confirmAppealDeposit);
         if (appeal.appealedRuling == finalRuling) {
+            // No need for SafeMath: collateral factors > 0 => both appeal deposits > totalFees
```
izqui: ~Unrelated Q: do we allow any of the collateral factors to be 0? I bet something funny could happen if that's allowed~

Edit: just continued reading and saw that you caught it already :D

contracts/court/Court.sol
```diff
@@ -1297,6 +1325,7 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
         // TODO: add reasonable limits for durations
         for (uint i = 0; i < _roundStateDurations.length; i++) {
             require(_roundStateDurations[i] > 0, ERROR_CONFIG_PERIOD_ZERO_TERMS);
+            require(_roundStateDurations[i] < MAX_STATE_DURATION, ERROR_CONFIG_PERIOD_MAX_TERMS);
```
izqui: Nit: why not add them both in the same `require` and change the revert reason to something more generic like invalid adj state duration

contracts/lib/HexSumTree.sol
```diff
@@ -160,10 +169,12 @@ library HexSumTree {
 
         // Throw out-of-bounds error if there are no items in the tree or the highest value being searched is greater than the total
         uint256 total = getRecentTotalAt(self, _time);
+        // No need for SafeMath: positive length of array already checked
```
izqui: Hyper nit: so you are saying 0 isn't a positive number 🧐

contracts/lib/JurorsTreeSortition.sol
```diff
@@ -74,14 +77,13 @@ library JurorsTreeSortition {
         returns (uint256 low, uint256 high)
     {
         uint256 totalActiveBalance = tree.getRecentTotalAt(_termId);
-        // TODO: SafeMath
-        low = _selectedJurors * totalActiveBalance / _roundRequestedJurors;
+        low = _selectedJurors.mul(totalActiveBalance) / _roundRequestedJurors;
```
izqui: This would crash anyway, but do we have a check that prohibits request a draft of 0 jurors?

contracts/subscriptions/CourtSubscriptions.sol
izqui: That's a good catch. Should we add an explicit comment in this function as to why it is needed in this particular case?

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -638,34 +640,34 @@ contract CourtSubscriptions is TimeHelpers, Initializable, IsContract, ISubscrip
...
-                require(_periods >= resumePeriods + delayedPeriods, ERROR_LOW_RESUME_PERIODS_PAYMENT);
+                require(_periods >= resumePeriods.add(delayedPeriods), ERROR_LOW_RESUME_PERIODS_PAYMENT);
 
-                // Note that there is no need to use SafeMath here since we already checked the number of given and resume periods
+                // No need for SafeMath: we already checked the number of given and resume periods
                 regularPeriods = _periods - resumePeriods - delayedPeriods;
                 // The new last period is computed including the current period
-                // Note that there is no need to use SafeMath here since the number of periods is at least one
-                newLastPeriodId = _currentPeriodId + _periods - 1;
+                // No need for SafeMath: (for subtraction since) the number of periods is at least one.
```
izqui:
```suggestion
                // No need for SafeMath for the subtraction: the number of periods is always at least one.
```

contracts/court/Court.sol
bingen: Good idea, so we save some error messages space.

contracts/lib/HexSumTree.sol
bingen: Yes, I am. I guess it depends on how you define "positive" but it's something quite accepted:
https://en.wikipedia.org/wiki/0#Mathematics
http://mathworld.wolfram.com/Zero.html
http://mathforum.org/library/drmath/view/58735.html
Anyway, this comment is in plain English, and I think it's clear enough, although that's why [I really like using more formal mathematics notation](https://github.com/aragon/aragon-court/pull/162#discussion_r333935708), to avoid ambiguities. :wink:

contracts/subscriptions/CourtSubscriptions.sol
bingen: I'm using this format (fixed "No need for SafeMath:" prefix and if needed operation clarification afterwards in parenthesis), and as [Facu suggested](https://github.com/aragon/aragon-court/pull/162#discussion_r333910194) I want to keep it as consistent as possible, so I'll just remove the "since" (which I forgot) from there.

contracts/lib/JurorsTreeSortition.sol
bingen: We had one for initial number of jurors, now I added it for the appeal step factor. I'll add a comment too.

PR #163, opened 2019-10-12
Court & Voting: compute weight only on commit

bingen:
It only makes sense to compute juror's weight on commit, not on
reveal. A juror that has not commited won't be able to reveal anyway.

For that reason, order of checks is changed on CRVoting: it makes more
sense and it's more efficient to check conditions locally first and
then if everything is ok go to get weight to Court.

:warning: Merge after #162 and change base to development.

### DIFF
--- a/contracts/court/Court.sol
+++ b/contracts/court/Court.sol
@@ -315,7 +315,10 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     ) public {
         // This seems reasonable enough, and this way we avoid using SafeMath for termDurarion
         require(_termDuration < ONE_YEAR, ERROR_TERM_DURATION_TOO_LONG);
-        require(_firstTermStartTime >= getTimestamp64() + _termDuration, ERROR_BAD_FIRST_TERM_START_TIME);
+        require(
+            _firstTermStartTime >= getTimestamp64() + _termDuration && _firstTermStartTime <= getTimestamp64() + 2 * ONE_YEAR,
+            ERROR_BAD_FIRST_TERM_START_TIME
+        );
 
         termDuration = _termDuration;
         jurorsRegistry = _jurorsRegistry;
@@ -642,8 +645,9 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
 
         // Distribute the collected tokens of the jurors that were slashed weighted by the winning jurors. Note that
         // we are penalizing jurors that refused intentionally their vote for the final round.
+        // Note that coherentJurors has to be > 0 because above it's required that _juror voted in favour (so at the very least there is one),
+        // therefore divisions below are safe.
         uint256 coherentJurors = round.coherentJurors;
-        require(coherentJurors > 0, ERROR_NO_COHERENT_JURORS);
         uint256 collectedTokens = round.collectedTokens;
         if (collectedTokens > 0) {
             jurorsRegistry.assignTokens(_juror, uint256(jurorState.weight).mul(collectedTokens) / coherentJurors);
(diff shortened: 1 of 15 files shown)


contracts/court/Court.sol
```diff
@@ -724,6 +724,20 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    * @return Weight of the requested juror for the requested dispute's round
+    */
+    function getVoterWeightToLeak(uint256 _voteId, address _voter) external only(address(voting)) ensureTerm returns (uint64) {
+        (uint256 disputeId, uint256 roundId) = _decodeVoteId(_voteId);
+        Dispute storage dispute = disputes[disputeId];
+        _checkAdjudicationState(dispute, roundId, AdjudicationState.Committing);
+        AdjudicationRound storage round = dispute.rounds[roundId];
+        return _getStoredJurorWeight(round, _voter);
+    }
+
```
facuspagnuolo: This one may not be necessary if we keep the cache in the compute method or we try to load a juror's weight when committing votes.

contracts/court/Court.sol
```diff
@@ -1412,13 +1421,14 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
...
+    * @dev Internal function to get the stored juror weight for a round. Note that the weight of a juror is:
+    *      - For a regular round: the number of times a juror was picked for the round round.
+    *      - For a final round: the proportion of juror's active state over the total, only set after the juror has voted.
     * @param _round Dispute round to calculate the juror's weight of
     * @param _juror Address of the juror to calculate the weight of
-    * @return Weight of the requested juror for the final round of the given dispute
+    * @return Weight of the requested juror for the given round
     */
-    function _getJurorWeightForRegularRound(AdjudicationRound storage _round, address _juror) internal view returns (uint64) {
+    function _getStoredJurorWeight(AdjudicationRound storage _round, address _juror) internal view returns (uint64) {
```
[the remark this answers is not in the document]
bingen: Yes, but the problem is that it's not exclusively for regular rounds anymore. For leaks and reveals it can be called for final rounds too.

contracts/voting/CRVoting.sol
```diff
@@ -269,6 +269,18 @@ contract CRVoting is Initializable, ICRVoting {
...
 
+    /**
+    * @dev Internal function to fetch and ensure the weight of a voter whose vote is being leaked a vote for a given vote instance
+    * @param _voteId ID of the vote instance to check the voter's weight of
+    * @param _voter Address of the voter whose vote is being leaked
+    * @return Weight of the voter whose vote is being leaked
+    */
+    function _ensureVoterWeightToLeak(uint256 _voteId, address _voter) internal returns (uint256) {
+        uint256 weight = uint256(owner.getVoterWeightToLeak(_voteId, _voter));
+        require(weight > 0, ERROR_REVEAL_DENIED_BY_OWNER);
```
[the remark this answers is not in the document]
bingen: Actually I would use the same error for the three of them (just `ERROR_DENIED_BY_OWNER`), but since it was already separated, I'll keep it this way.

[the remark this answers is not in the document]
bingen: Actually now I'm thinking that this should never happen, and the same applies to reveal: to be able to leak or reveal, juror must have commited first, and therefore should have `weight > 0`. So I'm removing this, ok?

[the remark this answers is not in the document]
bingen: Actually now I'm thinking that this should never happen, and the same applies to reveal: to be able to leak or reveal, juror must have commited first, and therefore should have `weight > 0`.  Only term needs to be checked. So I'm changing this.

contracts/court/Court.sol
```diff
@@ -1412,13 +1421,14 @@ contract Court is IJurorsRegistryOwner, ICRVotingOwner, ISubscriptionsOwner, Tim
     }
 
     /**
-    * @dev Internal function to get the juror weight for a regular round. Note that the weight of a juror for a regular
-    *      round is the number of times a juror was picked for the round round.
+    * @dev Internal function to get the stored juror weight for a round. Note that the weight of a juror is:
+    *      - For a regular round: the number of times a juror was picked for the round round.
+    *      - For a final round: the proportion of juror's active state over the total, only set after the juror has voted.
```
izqui:
```suggestion
    *      - For a final round: the relative active stake of a juror's state over the total active tokens, only set after the juror has voted.
```

contracts/court/Court.sol
bingen: Not sure if I'm getting this.

facuspagnuolo:
I'm sure I haven't reviewed the `ensureTermAndGetVoterWeightToReveal` and the same for commit, but I think these are bad names, too white box IMO. I'd change them for sth like `ensureVoterWeightToReveal` or so.

contracts/court/Court.sol
facuspagnuolo: Sorry that I'm answering this a bit late. I don't think we need an extra external method to get the voter's weight when trying to leak a vote we can use the same we use for committing
if we [this cache](https://github.com/aragon/aragon-court/pull/163/files#diff-d3682a59f2ba9e8e2b89bd9a2e2fa2b2L1248-L1253). 

In general, it is not good having separate domain contracts talking about each other's domain

bingen:
> I'm sure I haven't reviewed the `ensureTermAndGetVoterWeightToReveal` and the same for commit, but I think these are bad names, too white box IMO. I'd change them for sth like `ensureVoterWeightToReveal` or so.

Yes, sorry about that, but I merged it because @izqui reviewed it and we agreed off-line I would say so. I said I would do this change [here](https://github.com/aragon/aragon-court/pull/163#discussion_r334368970). Of course I didn't get into the details of the name, but that is the whole point of the change: now we don't ensure the voter weight on reveal, but we ensure the term and get the weight, and hence the name. Also, how come adding "TermAnd" in between make them more "white box". What do you mean with "white box"?

contracts/court/Court.sol
bingen: Sorry, I still don't get it. The cache was being used, with `_getStoredJurorWeight`. Besides, wasn't it the same we were doing with reveal? I don't get which domains you think were being mixed here.
Anyway, it doesn't matter anymore, because now leak check doesn't care about the weight (that comment was on an outdated version).

PR #164, opened 2019-10-12
Court & Registry: move final round computations to registry

bingen:
Avoid some calls to registry, reduce Court contract size a little bit.

:warning: Merge after #163 and change base to development.

### DIFF
--- a/contracts/registry/JurorsRegistry.sol
+++ b/contracts/registry/JurorsRegistry.sol
@@ -372,8 +372,7 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
     * @return Amount of active tokens for juror in the requested past term id
     */
     function activeBalanceOfAt(address _juror, uint64 _termId) external view returns (uint256) {
-        Juror storage juror = jurorsByAddress[_juror];
-        return _existsJuror(juror) ? tree.getItemAt(juror.id, _termId) : 0;
+        return _activeBalanceOfAt(_juror, _termId);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


contracts/registry/JurorsRegistry.sol
```diff
@@ -393,6 +392,47 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
...
+    * @dev Calculate a juror's active balance and the number of times that it contains the min active balance.
+    *      Used to get the juror weight for the final round. Note that for the final round the weight of
+    *      each juror is equal to the number of times the min active balance the juror has, multiplied by a precision
+    *      factor to deal with division rounding.
+    * @param _juror Address of the juror to calculate the balance info
+    * @param _termId Term querying the balance info
+    * @param _precision Multiplier to mitigate division rounding errors
+    * @return Weight of the requested juror for the final round of the given dispute
+    */
+    function getActiveBalanceInfo(address _juror, uint64 _termId, uint256 _precision)
```
facuspagnuolo: WDYT about `getActiveBalanceShareOfAt`?

contracts/registry/JurorsRegistry.sol
```diff
@@ -393,6 +392,47 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
...
+    *      Used to calculate the jurors number for final rounds at the current term.
+    * @param _termId Term querying that multiple
+    * @param _precision Multiplier to mitigate division rounding errors
+    * @return The number of times that the total active balance contains the min active balance.
+    *         Equals Jurors number for final rounds for the given term.
+    */
+    function getTotalMinActiveBalanceShares(uint64 _termId, uint256 _precision) external view returns (uint256) {
+        // Total active balance is guaranteed to never be greater than
+        // `2^64 * minActiveBalance / FINAL_ROUND_WEIGHT_PRECISION`. Thus, the
+        // jurors number for a final round will always fit in `uint64`
```
facuspagnuolo: I think we should move this comment back to the `Court` contract since we're doing the cast there actually

contracts/registry/JurorsRegistry.sol
bingen: The `Of` and `At` makes total sense, adding it. Adding the `Share` part IMO could mislead to thinking that we are only getting the relation to min active balance.

contracts/registry/JurorsRegistry.sol
bingen: Totally.

contracts/registry/JurorsRegistry.sol
```diff
@@ -393,6 +392,44 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
...
+    *      Used to get the juror weight for the final round. Note that for the final round the weight of
+    *      each juror is equal to the number of times the min active balance the juror has, multiplied by a precision
+    *      factor to deal with division rounding.
+    * @param _juror Address of the juror to calculate the balance info
+    * @param _termId Term querying the balance info
+    * @param _precision Multiplier to mitigate division rounding errors
+    * @return Weight of the requested juror for the final round of the given dispute
+    */
+    function getActiveBalanceInfoOfAt(address _juror, uint64 _termId, uint256 _precision)
+        external view returns (uint256 activeBalance, uint256 shares)
```
izqui: `shares` is a bit of a weird word here IMO. How about `slots` or `minActiveBalanceMultiple`?

And why not just do the division at the Court.sol level? Is it not part of the Court's state anymore?

contracts/registry/JurorsRegistry.sol
```diff
@@ -393,6 +392,44 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
         return jurorsByAddress[_juror].id;
     }
 
+    /**
+    * @dev Calculate the number of times that the total active balance contains the min active balance.
```
izqui:
```suggestion
    * @dev Calculate the number of times that the total active balance contains the min active balance (multiplied by inputted precision).
```

contracts/registry/JurorsRegistry.sol
```diff
@@ -393,6 +392,44 @@ contract JurorsRegistry is Initializable, IsContract, IJurorsRegistry, ERC900, A
...
+    * @param _precision Multiplier to mitigate division rounding errors
+    * @return The number of times that the total active balance contains the min active balance.
+    *         Equals Jurors number for final rounds for the given term.
+    */
+    function getTotalMinActiveBalanceShares(uint64 _termId, uint256 _precision) external view returns (uint256) {
+        return _precision.mul(_totalActiveBalanceAt(_termId)) / minActiveBalance;
+    }
+
+    /**
+    * @dev Calculate a juror's active balance and the number of times that it contains the min active balance.
```
izqui:
```suggestion
    * @dev Calculate a juror's active balance and the number of times that it contains the min active balance (multiplied by precision)
```

contracts/registry/JurorsRegistry.sol
bingen: I thought of both `slots` and `multiple`, none of them convinced me fully. Switching to `slots` then.

Issue #43, opened 2019-05-21
Contracts: Stack too deep issues


bingen:
Fix stack too deep issues in Court contract:

- `draftAdjudicationRound`
- `settleRoundSlashing`

Once it's fixed, optimize `SSTORE` and `SLOAD` use.

See comments [here](https://github.com/aragon/aragon-court/pull/41#discussion_r285199861) and [here](https://github.com/aragon/aragon-court/pull/41#discussion_r285199452) for more context.

izqui:
Do we want to get this in for the freeze?

PR #169, opened 2019-10-17
Court: Split terms management into a separate contract

facuspagnuolo:
This PR introduces a new module called `CourtClock` for now in charge of handling all the terms-related behavior. The main idea is to split the terms management from the rest of the `Court` behavior. This allows us to fix the EIP-170 limit issue. 

The main con of doing this is that in order to access term-related information from the `Court` contract we need to perform 2 calls now: one to fetch the `CourtClock` module and another to perform the desired query. I've been discussing this with @izqui and we came up with the idea of having the `Controller` inheriting from this module instead. The problem with this is that we won't be able to change it if needed. 

See below some gas costs measures I got by running [these tests](1184c4ab26726d61a4f2432e9e3eabe8db715d7e) for different implementations:

-------

Here are the gas costs comparisons:

**Last commit when EIP-170 was not a problem** (a5cc9cce161416aed3f3498c68e44c5dcb2ec6cb)

| Function            | Without heartbeat | With heartbeat |
|---------------------|-------------------|---|
|       createDispute |           214,767 | 273,864 |
|               draft |           348,444 | invalid |
|              commit |            58,939 | 103,217 |
|              reveal |            86,790 | 131,068 |
|        createAppeal |            56,961 | 101,241 |
|       confirmAppeal |           178,189 | 222,469 |
|       executeRuling |            50,554 | 94,833 |
|     settlePenalties |           150,623 | 187,940 |
|        settleReward |            66,244 | 110,523 |
| settleAppealDeposit |            66,449 | 111,002 |

**Clock as a separate module**

| Function            | Without heartbeat | With heartbeat |
|---------------------|-------------------|---|
|       createDispute |           233,765 | 284,156 |
|               draft |           370,214 | invalid |
|              commit |            68,154 | 103,790 |
|              reveal |            94,663 | 130,299 |
|        createAppeal |            66,034 | 101,670 |
|       confirmAppeal |           191,812 | 227,448 |
|       executeRuling |            57,383 | 93,019 |
|     settlePenalties |           179,073 | 214,709 |
|        settleReward |            70,740 | 70,740 |
| settleAppealDeposit |            69,131 | 68,861 |

**Clock as part of the Controller**

| Function            | Without heartbeat | With heartbeat |
|---------------------|-------------------|---|
|       createDispute |           234,405 | 289,762 |
|               draft |           368,000 | invalid |
|              commit |           101,238 | 126,163 |
|              reveal |            97,138 | 152,676 |
|        createAppeal |            68,177 | 123,715 |
|       confirmAppeal |           194,440 | 249,978 |
|       executeRuling |            59,450 | 114,988 |
|     settlePenalties |           177,860 | 240,062 |
|        settleReward |            71,278 | 71,278 |
| settleAppealDeposit |            69,343 | 69,613 |

**Config as part of the Controller**

| Function            | Without heartbeat | With heartbeat |
|---------------------|-------------------|---|
|       createDispute |           216,417 | 272,902 |
|               draft |           394,946 | invalid |
|              commit |            90,021 | 146,687 |
|              reveal |           104,426 | 161,092 |
|        createAppeal |            85,268 | 141,934 |
|       confirmAppeal |           190,988 | 247,654 |
|       executeRuling |            66,264 | 122,930 |
|     settlePenalties |           206,879 | 270,196 |
|        settleReward |            83,216 | 83,216 |
| settleAppealDeposit |            79,415 | 79,685 |


test/court/court-disputes.js
```diff
@@ -42,8 +48,6 @@ contract('Court', ([_, sender]) => {
           })
 
           it('creates a new dispute', async () => {
-            // move forward to the term before the desired start one for the dispute
-            await courtHelper.setTerm(draftTermId - 1)
```
bingen: Why is this "flaky"? It just makes sure that the dispute is created for the desired `draftTermId`. When this was added, we used to have: 
```
const receipt = await court.createDispute(arbitrable.address, possibleRulings, draftTermId, { from: sender })
```
Now we don't pass the `draftTermId` anymore and instead disputes are created in the next term, so this added line makes sure that the test is behaving exactly the same as before.
Besides, if for any reason you change `draftTermId` to `3` in line 40 just above, without this line the test will fail, while with it it will work. So this line also makes the test more robust.

bingen: Btw, I realize I already told you this here:
https://github.com/aragon/aragon-court/pull/150#discussion_r332692561
(but you didn't reply)

facuspagnuolo: I've been having some weird behavior and I was playing with it a little bit, suddenly it started working again but apparently I push that stuff by mistake, it wasn't part of this PR. Thanks!

bingen: Yes, the change of not passing the `draftTermId` now is kind of confusing until we get used to it :wink:

contracts/court/Court.sol
```diff
@@ -436,7 +373,7 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
     * @param _roundId Identification number of the dispute round being appealed
     * @param _ruling Ruling appealing a dispute round in favor of
     */
-    function createAppeal(uint256 _disputeId, uint256 _roundId, uint8 _ruling) external ensureTerm {
```
bingen: Shouldn't all of these be replaced by `_ensureTermId()` call?

contracts/court/Court.sol
```diff
@@ -1285,17 +1135,15 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
...
-        // keep the known config
-        require(termId == 0 || _fromTermId > termId + 1, ERROR_TOO_OLD_TERM);
+        // If the current term is not zero, changes must be scheduled at least 2 terms in the future.
+        // This way we can ensure that disputes scheduled for the next term won't have their config changed.
+        require(_currentTermId == ZERO_TERM_ID || _fromTermId > _currentTermId + 1, ERROR_TOO_OLD_TERM);
 
         require(_appealCollateralParams[0] > 0 && _appealCollateralParams[1] > 0, ERROR_ZERO_COLLATERAL_FACTOR);
+
         // Make sure the given penalty pct is not greater than 100%
-        uint16 _penaltyPct = _pcts[0];
```
bingen: Why this change? Does it have any advantage? In that case, shouldn't we do it for similar params like `_firstRoundJurorsNumber`

izqui:
On average this gives us a hit of ~10k gas per call, I think for such a hot code path we should strive to optimize as much as possible. I still consider CourtClock a core part of the protocol, and if a bug is found there we are going to be so fucked anyway.

bingen:
Yes, I agree. I hadn't seen those numbers. For some calls it's even worse, like `createDispute` or `draft`, that are around ~20k, or `heartbeat` that is ~40k. Plus we already added some overhead for some calls [here](https://github.com/aragon/aragon-court/pull/151#pullrequestreview-296707719).

contracts/court/CourtClock.sol
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+    * @dev Internal function to compute the randomness that will be used to draft jurors for the given term. This
+    *      function assumes the given term exists. To determine the randomness factor for a term we use the hash of a
+    *      block number that is set once the term has started to ensure it cannot be known beforehand. Note that the
+    *      hash function being used only works for the 256 most recent block numbers.
+    * @param _term Term to compute the randomness of
+    * @return Randomness computed for the given term
+    */
+    function _getTermRandomness(Term storage _term) internal view returns (bytes32) {
```
izqui: Super nit regarding naming: when I read this function I assumed that given that it took a `Term` as an argument, it was just going to return `term.randomnessBN`. Maybe add a stronger connotation that it is going to get it from a blockhash or even compute it directly?

contracts/court/Court.sol
```diff
@@ -723,64 +660,36 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
...
-    * @return Randomness of the requested term
-    */
-    function getTermRandomness(uint64 _termId) external view returns (bytes32) {
-        require(_termId <= termId, ERROR_TERM_DOES_NOT_EXIST);
-        Term storage term = terms[_termId];
-        return _getTermRandomness(term);
+            // Pay heartbeat fees to the caller of this function
+            if (totalFee > 0) {
+                IAccounting accounting = _accounting();
+                accounting.assign(config.token, msg.sender, totalFee);
```
izqui: We should handle when the feeToken changes during heartbeats, as it stands right now it would break the accounting (we'd be issuing payments in a token with amounts of another). An easy way to handle it would be by reverting, not allowing processing multiple term transitions if the feeToken is changed

contracts/court/Court.sol
facuspagnuolo: Yes, but given that all these are using `_checkAdjudicationState` the term actually being ensured there, we can split these tho to make it explicit

bingen: Or maybe just add a comment?

contracts/court/Court.sol
```diff
@@ -391,22 +333,18 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
...
+        // Ensure draft term randomness can be computed for the current block number
         AdjudicationRound storage round = dispute.rounds[dispute.rounds.length - 1];
-        uint64 requestedDraftTermId = round.draftTermId;
-        uint64 currentTermId = termId;
-        require(requestedDraftTermId <= currentTermId, ERROR_ROUND_NOT_DRAFT_TERM);
-
-        // Ensure current term randomness can be ensured for the current block number
-        Term storage draftTerm = terms[currentTermId];
-        _ensureTermRandomness(draftTerm);
+        bytes32 draftTermRandomness = clock.ensureTermRandomness(round.draftTermId);
```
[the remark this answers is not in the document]
facuspagnuolo: Had to resign one because of a stack too deep issue

contracts/court/Court.sol
```diff
@@ -75,6 +62,9 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
     string private constant ERROR_NO_COHERENT_JURORS = "CT_NO_COHERENT_JURORS";
     string private constant ERROR_ROUND_APPEAL_ALREADY_SETTLED = "CT_APPEAL_ALREADY_SETTLED";
 
+    // Initial term to start the Court, disputes are not allowed during this term. It can be used to active jurors
+    uint64 internal constant ZERO_TERM_ID = 0;
```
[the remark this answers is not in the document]
facuspagnuolo: WDYT about moving these constants to a reusable contract? Something similar to what we are doing in aragonOS (see a6d1a90863a6b2e964057e09b6bd815b526bc2bb)

facuspagnuolo:
@bingen @izqui addressed all your comments, feel free to have another look at it

contracts/court/Court.sol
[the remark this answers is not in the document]
bingen: Moving these constants to a reusable contract would be the minimum IMO, but I insist I don't think we need it as it's just the first element in the array, we do it all the time:
Should this `0` and this `1` here be moved to a constant too?
https://github.com/aragon/aragon-court/blob/terms_split/contracts/court/Court.sol#L251
https://github.com/aragon/aragon-court/blob/terms_split/contracts/court/Court.sol#L1161
https://github.com/aragon/aragon-court/blob/terms_split/contracts/court/Court.sol#L1223
Similarly:
https://github.com/aragon/aragon-court/blob/terms_split/contracts/voting/CRVoting.sol#L225
https://github.com/aragon/aragon-court/blob/terms_split/contracts/registry/JurorsRegistry.sol#L261
https://github.com/aragon/aragon-court/blob/terms_split/contracts/registry/JurorsRegistry.sol#L699
https://github.com/aragon/aragon-court/blob/terms_split/contracts/controller/Controller.sol#L181
etc

contracts/court/Court.sol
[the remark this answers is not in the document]
bingen: Oh, ok, it would be nice to add a `TODO` comment, so we can easily find these cases when we tackle #43

contracts/court/Court.sol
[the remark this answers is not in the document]
bingen: As I said before, if you insist to keep those constants, `CourtClock` constructor should have `termID = ZERO_TERM_ID;` (equivalent to "heartbeat" to that zero term).

bingen:
Hm, I'm not sure, tbh. It seems a little bit convoluted in my opinion: the fact that `Controller` has `Court` (which is `Controlled`) as an owner (`IClockOwner`) looks a little bit weird to me.
It would be nice to run gas tests again and see how much we gained with these last changes.
I checked `Court` bytecode size and it's at `24461`, so a little bit tight.

facuspagnuolo:
@bingen I honestly don't like how this is looking as well to be honest, but I don't see another way of inheriting the clock from the controller without doing that. Regarding the gas costs, note that I updated the description with a new table

contracts/court/Court.sol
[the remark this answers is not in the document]
facuspagnuolo: > Should this 0 and this 1 here be moved to a constant too?

Yes

> Similarly

No

> if you insist to keep those constants, CourtClock constructor should have termID = ZERO_TERM_ID

Sure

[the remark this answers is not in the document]
bingen: > No

Could you elaborate a little bit more, please?

contracts/court/Court.sol
```diff
@@ -1759,11 +1228,10 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
 
         // Store or update drafted jurors' weight
         for (uint256 i = 0; i < outputLength; i++) {
-            address juror = jurors[i];
```
bingen: Is this a stack too deep issue?

contracts/controller/config/CourtConfig.sol
```diff
@@ -0,0 +1,270 @@
...
+import "./IConfig.sol";
+import "./CourtConfigData.sol";
+import "../clock/CourtClock.sol";
+import "../../lib/PctHelpers.sol";
+
+
+contract CourtConfig is IConfig, CourtConfigData, CourtClock {
+    using PctHelpers for uint256;
+
+    string private constant ERROR_TOO_OLD_TERM = "CT_TOO_OLD_TERM";
```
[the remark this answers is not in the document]
facuspagnuolo: Simply forgot about this one :)

contracts/controller/Controller.sol
```diff
@@ -73,14 +78,111 @@ contract Controller is IsContract {
...
+    * @param _maxRequestedTransitions Max number of term transitions allowed by the sender
+    * @return currentTermId Identification number of the term id after executing the heartbeat transitions
     */
-    constructor(address _fundsGovernor, address _configGovernor, address _modulesGovernor) public {
-        _setFundsGovernor(_fundsGovernor);
-        _setConfigGovernor(_configGovernor);
-        _setModulesGovernor(_modulesGovernor);
+    function heartbeat(uint64 _maxRequestedTransitions) external returns (uint64) {
+        (uint64 previousTermId, uint64 currentTermId) = _heartbeat(_maxRequestedTransitions);
+        _updateTermsConfig(previousTermId, currentTermId);
```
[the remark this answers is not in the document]
facuspagnuolo: Yea, I did this way to solve it quickly but I see a way we can improve it

contracts/court/Court.sol
facuspagnuolo: yup, added a comment

contracts/controller/clock/CourtClock.sol
```diff
@@ -66,6 +66,24 @@ contract CourtClock is IClock, TimeHelpers {
...
+    * @notice Transition up to `_maxRequestedTransitions` terms
+    * @param _maxRequestedTransitions Max number of term transitions allowed by the sender
+    * @return currentTermId Identification number of the term id after executing the heartbeat transitions
+    */
+    function heartbeat(uint64 _maxRequestedTransitions) external returns (uint64) {
+        return _heartbeat(_maxRequestedTransitions);
+    }
+
+    /**
+    * @notice Ensure that the current term of the Court is up-to-date. If the Court is outdated by more than one term, the heartbeat function
```
bingen: To be precise I would say here `MAX_AUTO_TERM_TRANSITIONS_ALLOWED` instead of `one`.

contracts/controller/clock/CourtClock.sol
```diff
@@ -190,8 +205,26 @@ contract CourtClock is IClock, TimeHelpers {
...
+        return currentTermId;
+    }
+
+    /**
+    * @dev Internal function to notify when a term has been transitioned
+    * @param _currentTermId Identification number of the new current term that has been transitioned
+    */
+    function _onTermTransitioned(uint64 _currentTermId) internal {
+        // solium-disable-previous-line no-empty-blocks
+        // This function must be override to provide custom behavior
```
bingen: :+1:
I think it's "overridden".

contracts/court/Court.sol
```diff
@@ -27,15 +27,6 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
     // Authorization-related error messages
```
izqui: Crazy how clean it has been to move the config, tells a lot about the architecture :)

contracts/controller/config/CourtConfigData.sol
```diff
@@ -0,0 +1,32 @@
+pragma solidity ^0.5.8;
+
+import "@aragon/os/contracts/lib/token/ERC20.sol";
+
+
+contract CourtConfigData {
+    struct Config {
```
izqui: Clean af

contracts/controller/Controller.sol
```diff
@@ -270,6 +375,14 @@ contract Controller is IsContract {
         emit ModuleSet(_id, _addr);
     }
 
+    /**
+    * @dev Internal function to notify when a term has been transitioned
+    * @param _currentTermId Identification number of the new current term that has been transitioned
+    */
+    function _onTermTransitioned(uint64 _currentTermId) internal {
```
izqui: Beautiful abstraction

PR #170, opened 2019-10-17
Court: Implement modifiers to ensure dispute and round existence

### LINKED ISSUE #62: Court: Add checks to ensure dispute and round exist
In most functions we are doing:
```
```
Since this is a mapping, should we check that the dispute actually exists? (Downside being adding more code, of course)

The same for:
```
        AdjudicationRound storage round = dispute.rounds[_roundId];
```

In a lot of them we are then doing too:
```
        AdjudicationRound storage round = dispute.rounds[dispute.rounds.length - 1];
```
Which would fail because rounds length is zero, but it would be nice to have a proper error message.

### LINKED ISSUE #158: Court: restore `disputeExists` and `roundExists` modifiers
It should just be `git revert 220d664`.
Commit 220d664 was a temporary solution to reduce contract size, allow CI to pass and to not block config PR, but it's not the accepted consensus.
Once another strategy is applied and the contract size is reduced again, that commit can be undone and modifiers restored.

facuspagnuolo:
**MERGE AFTER #169**

Fixes #62 
Fixes #158


contracts/court/Court.sol
```diff
@@ -629,8 +615,7 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
     * @return Weight of the requested juror for the requested dispute's round
     */
     function ensureTermAndGetVoterWeightToCommit(uint256 _voteId, address _voter) external onlyVoting returns (uint64) {
-        (uint256 disputeId, uint256 roundId) = _decodeVoteId(_voteId);
-        Dispute storage dispute = disputes[disputeId];
+        (Dispute storage dispute, uint256 roundId) = _decodeVoteId(_voteId);
```
bingen: :+1:

contracts/court/Court.sol
```diff
@@ -953,7 +928,6 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
     */
     function _checkAdjudicationState(Dispute storage _dispute, uint256 _roundId, AdjudicationState _state) internal {
         uint64 termId = _ensureTermId();
-        require(_roundId < _dispute.rounds.length, ERROR_ROUND_DOES_NOT_EXIST);
```
bingen: I would add a comment to this function saying that it assumes dispute and round exist.

PR #173, opened 2019-10-17
Court: Rename voting methods

facuspagnuolo:
**MERGE AFTER #172**

Following [this](https://github.com/aragon/aragon-court/pull/163) discussion, I propose the following changes:

1. Renaming `ensureTermAndGetVoterWeightToX` methods to simply `ensureVoterWeightToX`:
There are several reasons why I think we should use another name:
- We should try not to expose details about what's happening when computing a voter's weight. In the same way, we don't have a `ensureTermAndSettlePenalties` and so on. 
- Based on a previous discussion we had about when to use `ensure` as a prefix, we decided to use it when we are writing a value if certain conditions are not met. In this case, I think we could leverage it as well since we are calculating the voter's weight for final rounds.
- Using the word `and` in a function name to describe multiple behaviors is again white-box and dangerous since we cannot rely on a list forever. What if we have to add a new step to this function, are we going to concatenate it to the function name? 

2. Rename `_getStoredJurorWeight` to `_getJurorWeight`: 
- 90% of the getters are actually reading sth from storage, don't see why being inconsistent for this case.

3. Drop `ensureTermToLeak` method: 
- IMO this is mixing domains: why the `Court` should know sth about leaking votes? This is particular to the voting protocol the voting module is using. Note that the other way around, it isn't crazy having the voting module asking if a voter can still commit votes.
- By simply adding a [cache](https://github.com/aragon/aragon-court/compare/registry_setters...rename_voting_methods?expand=1#diff-d3682a59f2ba9e8e2b89bd9a2e2fa2b2R1055-R1057) to the `_computeJurorWeightForFinalRound` method we can avoid having this entry-point


bingen:
I'll review the code tomorrow, but some comments in advance:

> Based on a previous discussion we had about when to use ensure as a prefix, we decided to use it when we are writing a value if certain conditions are not met. In this case, I think we could leverage it as well since we are calculating the voter's weight for final rounds.

- First thing, for reveal and leak we are not calculating anything.
- We said that we would use ensure, as opposed to check, when instead of only checking if a value met a condition, we were also "forcing" it to meet it. The example is term: you can only check if term is up to date, or advance terms to make sure it is. We never do that with weights, obviously.

> Using the word and in a function name to describe multiple behaviors is again white-box

What do you mean with white-box? (I asked in that previous discussion, but didn't get a response)

>IMO this is mixing domains: why the Court should know sth about leaking votes? This is particular to the voting protocol the voting module is using. Note that the other way around, it isn't crazy having the voting module asking if a voter can still commit votes.

Didn't we already have `getVoterWeightToCommit`(https://github.com/aragon/aragon-court/blob/ead72067064e239a05c535c08949ce8d5a5d21d3/contracts/court/Court.sol#L720) and `getVoterWeightToReveal`(https://github.com/aragon/aragon-court/blob/ead72067064e239a05c535c08949ce8d5a5d21d3/contracts/court/Court.sol#L733)
What's the difference? Why can `Court` know about "commit" or "reveal" but not "leak"?

> By simply adding a cache to the `_computeJurorWeightForFinalRound` method we can avoid having this entry-point

Well, part of the point of the change was to avoid that unnecessary check. Now we only compute the weight on commit, and that only happens once, so there's no need to check if it has been already done.

facuspagnuolo:
Thanks for your thoughts, answering your questions below

> First thing, for reveal and leak we are not calculating anything.

True, but we are doing for committing, and I'd rather be consistent.

> We said that we would use ensure, as opposed to check, when instead of only checking if a value met a condition, we were also "forcing" it to meet it. 

Again, I'd prefer being consistent. I think this is a particular case for the prefix `ensure` we can extend for this case as well.

> What do you mean with white-box? (I asked in that previous discussion, but didn't get a response)

This is related to my first point: "We should try not to expose details about what's happening when computing a voter's weight. In the same way, we don't have a ensureTermAndSettlePenalties and so on". I.e. exposing details of what's being done inside of a function, opposite of encapsulation or black-boxing concept.

> Didn't we already have getVoterWeightToCommit and getVoterWeightToReveal
What's the difference? Why can Court know about "commit" or "reveal" but not "leak"?

Because commit and reveal are part of the court protocol, leak is part of the voting one

> Well, part of the point of the change was to avoid that unnecessary check. Now we only compute the weight on commit, and that only happens once, so there's no need to check if it has been already done.

If we can avoid having repeated code, another entry point, and reducing the code surface, by only adding a check, a simple syntax we commonly use to cache things, I don't see why this case would be seen as wrong. Again, we are mixing concepts here

bingen:
Before getting into the details I want to say in advance that I don't car about the names, put whatever you like, but I do about the concepts and arguments discussed here, so:

> True, but we are doing for committing, and I'd rather be consistent.

What do you mean? Putting names to reveal and leak related functions that are consistent with commit one, even if what the functions do is different? This doesn't make any sense to me. I mean, for consistency you could rename them all to `ensureVoterWeightToHearbeat`, `ensureVoterWeightToAppeal`, etc, but I don't see the point.

> Again, I'd prefer being consistent. I think this is a particular case for the prefix ensure we can extend for this case as well.

Again, extending the prefix `ensure` to functions that don't "ensure" anything doesn't make sense to me.

But I insist, if you really like it better, go ahead with it.

> ... opposite of encapsulation or black-boxing concept.

Oh, ok, now I get it, thanks for the clarification. I just disagree that using `And` in a name has that effect, to me it just makes it more explicit.
Besides, we have used it before:
https://github.com/aragon/aragonOS/blob/next/contracts/ens/ENSSubdomainRegistrar.sol#L64
https://github.com/aragon/aragonOS/blob/next/contracts/common/SafeERC20.sol#L17

> Because commit and reveal are part of the court protocol, leak is part of the voting one

What?? Who says so? `CRVoting` is named for "Commit-Reveal Voting". And the Court could definitely use another kind of secret Voting mechanism (or even not hide it, although that would be bad for jurors incentives).

>If we can avoid having repeated code, another entry point, and reducing the code surface, by only adding a check, a simple syntax we commonly use to cache things, I don't see why this case would be seen as wrong. Again, we are mixing concepts here

Well, in my opinion mixing concepts is asking for "voter weight to commit" when you really want to leak and you don't care at all about the weight, because you know that the vote was previously committed (otherwise you can't leak).
About entry points, you could have just one, with a huge params array and a switch case to route actions, but having an entry point for each different action seems quite natural to me.

facuspagnuolo:
> Before getting into the details I want to say in advance that I don't care about the names that much, put whatever you like

Cool, let's then not waste time reviewing this PR first when there a bunch of others more important

> but I do about the concepts and arguments discussed here, so:

Let's have a synchronous conversation then, we can jump on a call and paste the conclusion afterward so we don't leave the thread unconcluded.

bingen:
Cool, sounds good. Just let me add a question so we don't forget when we discuss, about the `And` in function names: would the same apply for `Or`? We have a close example:
https://github.com/aragon/aragon-court/blob/development/contracts/registry/JurorsRegistry.sol#L264

bingen:
> Let's have a synchronous conversation then

That discussion didn't happen, I didn't review the code (just discussed based on the PR description), @izqui either as far as I know, so nobody approved it. I don't think this PR should have been merged yet.

PR #174, opened 2019-10-18
Accounting: Add unprotected withdraw method

### LINKED ISSUE #168: CourtAccounting: Unprotect withdraw
Tokens can only go to owner, so we can have a daemon that forces withdraws for users so they don’t have to do anything manually.

facuspagnuolo:
**MERGE AFTER #172**

Fixes #168


contracts/accounting/CourtAccounting.sol
```diff
@@ -59,11 +59,22 @@ contract CourtAccounting is ControlledRecoverable, IAccounting {
...
+        _withdraw(_token, msg.sender, _to, _amount, newBalance);
+    }
 
-        require(_token.safeTransfer(_to, _amount), ERROR_WITHDRAW_FAILED);
+    /**
+    * @notice Withdraw all the tokens from `_to` to themself
+    * @param _token ERC20 token to be withdrawn
+    * @param _to Address of the recipient that will receive their tokens
+    */
+    function withdrawAll(ERC20 _token, address _to) external {
```
izqui: What's the rationale behind not allowing to set a custom amount for unprotected withdrawals?

bingen: Well, not sure (will let Facu respond), but the way I see it, if the owner is doing it, the regular withdraw is available allowing to choose the amount. If a 3rd party, like a bot, is doing it, why would it want to choose an amount? Are we thinking of any other use case that may need that amount param?

contracts/accounting/CourtAccounting.sol
```diff
@@ -59,11 +59,22 @@ contract CourtAccounting is ControlledRecoverable, IAccounting {
         require(_amount > 0, ERROR_WITHDRAW_AMOUNT_ZERO);
         require(balance >= _amount, ERROR_WITHDRAW_INVALID_AMOUNT);
 
-        address tokenAddress = address(_token);
-        balances[tokenAddress][msg.sender] = balance.sub(_amount);
-        emit Withdraw(_token, msg.sender, _to, _amount);
+        // No need for SafeMath: checked above
+        uint256 newBalance = balance - _amount;
+        _withdraw(_token, msg.sender, _to, _amount, newBalance);
```
bingen: To me it feels quite weird that we are passing `newBalance` as a parameter instead of calculating inside the internal function. To start with, it makes `_witdraw` less safe. For the same reason I would move both `require`'s inside it too.

contracts/accounting/CourtAccounting.sol
facuspagnuolo: Yea, I was mostly driven by Bingen's comment about having a daemon process using this. We can add it tho, no strong opinions

contracts/accounting/CourtAccounting.sol
facuspagnuolo: Yea I agree it is weird, I was trying to share as much code as possible while trying to have the minimum number of sloads as well

facuspagnuolo:
@bingen I think we can do something to improve that experience, I created #184 to tackle that

PR #177, opened 2019-10-18
Court: Make draft batches size fixed.

### LINKED ISSUE #167: Court: Fix drafter's ability to impact draft randomness
Due to the drafter being able to set the draft batch size, they can have multiple drafting options to choose from, so the draft is no longer random (or the drafter has at least 3x more chances than anyone else).

bingen:
And allow to change it by governor. To prevent probability
manipulation when drafting.

Fixes #167.


contracts/court/Court.sol
```diff
@@ -275,6 +281,7 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
         Controller _controller,
         uint64 _termDuration,
         uint64 _firstTermStartTime,
+        uint64 _maxJurorsToBeDraftedPerBatch,
```
facuspagnuolo: WDYT about `maxJurorsPerDraftBatch` ?

contracts/court/Court.sol
```diff
@@ -217,6 +221,7 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
 
     event NewTerm(uint64 termId, address indexed heartbeatSender);
     event NewCourtConfig(uint64 fromTermId, uint64 courtConfigId);
+    event MaxJurorsToBeDraftedPerBatchSet(uint64 oldMaxJurorsToBeDraftedPerBatch, uint64 newMaxJurorsToBeDraftedPerBatch);
```
facuspagnuolo: I think in the rest of the modules we are using `previous` and `current` naming for these

contracts/court/Court.sol
```diff
@@ -1254,6 +1271,16 @@ contract Court is TimeHelpers, Controlled, ControlledRecoverable, IJurorsRegistr
...
     }
 
+    /**
+    * @dev Sets the global configuration for the max number of jurors to be drafted per batch
+    * @param _maxJurorsToBeDraftedPerBatch Max number of jurors to be drafted per batch
+    */
+    function _setMaxJurorsToBeDraftedPerBatch(uint64 _maxJurorsToBeDraftedPerBatch) internal {
+        require(_maxJurorsToBeDraftedPerBatch > 0, ERROR_BAD_MAX_DRAFT_BATCH_SIZE);
+        emit MaxJurorsToBeDraftedPerBatchSet(maxJurorsToBeDraftedPerBatch, _maxJurorsToBeDraftedPerBatch);
+        maxJurorsToBeDraftedPerBatch = _maxJurorsToBeDraftedPerBatch;
```
facuspagnuolo: Just curious, any reason for not making it part of the court config?

test/court/court-draft-gas.js
```diff
@@ -0,0 +1,38 @@
...
+  let courtHelper, court, disputeId
+
+  const jurors = [
+    { address: juror500,  initialActiveBalance: bigExp(500,  21) },
+    { address: juror1000, initialActiveBalance: bigExp(1000, 21) },
+    { address: juror1500, initialActiveBalance: bigExp(1500, 21) },
+    { address: juror2000, initialActiveBalance: bigExp(2000, 21) },
+  ]
+
+  const deployAndcreateDispute = async (jurorsNumber) => {
```
facuspagnuolo: Let’s use `maxJurorsPerDraftBatch` here

test/court/court-draft-gas.js
```diff
@@ -0,0 +1,38 @@
...
+    const draftTermId = bn(1)
+    courtHelper = buildHelper()
+    court = await courtHelper.deploy({ maxJurorsToBeDraftedPerBatch: jurorsNumber, firstRoundJurorsNumber: jurorsNumber })
+    await courtHelper.activate(jurors)
+    disputeId = await courtHelper.dispute({ draftTermId })
+    await courtHelper.passRealTerms(1)
+  }
+
+  it('measures gas', async () => {
+    for (let j of [ 3, 5, 10, 50, 100 ]) {
```
facuspagnuolo: Same here instead of `j`

test/court/court-draft-gas.js
```diff
@@ -0,0 +1,38 @@
...
+      await deployAndcreateDispute(j)
+      const gas = await court.draft.estimateGas(disputeId)
+      console.log(`Number of jurors: ${j}, gas: ${gas}, ratio: ${Math.round(gas / j)}`)
+      const receipt = await court.draft(disputeId)
+      const logs = decodeEventsOfType(receipt, JurorsRegistry.abi, 'JurorDrafted')
+      assertAmountOfEvents({ logs }, 'JurorDrafted', j)
```
facuspagnuolo: We should have an object with the expected gas cost per value of `maxJurorsPerDraftBatch` instead and make a proper assertion here, no one will notice this logging otherwise

test/court/court-draft.js
```diff
@@ -252,29 +252,45 @@ contract('Court', ([_, disputer, drafter, juror500, juror1000, juror1500, juror2
...
+          })
+
+          itDraftsRequestedRoundInOneBatch(term, maxJurorsToBeDraftedPerBatch)
         })
 
         context('when drafting more than the requested jurors', () => {
-          const jurorsToBeDrafted = firstRoundJurorsNumber * 2
+          const maxJurorsToBeDraftedPerBatch = firstRoundJurorsNumber * 2
+
+          beforeEach('create court and dispute', async () => {
```
facuspagnuolo:
```suggestion
          beforeEach('set max number of jurors to be drafted per batch', async () => {
```

test/court/court-draft.js
```diff
@@ -252,29 +252,45 @@ contract('Court', ([_, disputer, drafter, juror500, juror1000, juror1500, juror2
...
+            itDraftsRequestedRoundInMultipleBatches(term, firstRoundJurorsNumber, batches, maxJurorsToBeDraftedPerBatch)
           })
         })
 
         context('when half amount of the requested jurors', () => {
-          const jurorsToBeDrafted = Math.floor(firstRoundJurorsNumber / 2)
+          const maxJurorsToBeDraftedPerBatch = Math.floor(firstRoundJurorsNumber / 2)
 
-          itDraftsRequestedRoundInOneBatch(term, jurorsToBeDrafted)
+          beforeEach('create court and dispute', async () => {
```
facuspagnuolo:
```suggestion
          beforeEach('set max number of jurors to be drafted per batch', async () => {
```

test/court/court-draft.js
```diff
@@ -252,29 +252,45 @@ contract('Court', ([_, disputer, drafter, juror500, juror1000, juror1500, juror2
...
+
+            itDraftsRequestedRoundInOneBatch(term, maxJurorsToBeDraftedPerBatch)
           })
 
           context('when drafting in multiple batches', () => {
-            const batches = 2, jurorsPerBatch = 4
+            const batches = 2, maxJurorsToBeDraftedPerBatch = 4
 
-            itDraftsRequestedRoundInMultipleBatches(term, jurorsToBeDrafted, batches, jurorsPerBatch)
+            beforeEach('create court and dispute', async () => {
```
facuspagnuolo:
```suggestion
            beforeEach('set max number of jurors to be drafted per batch', async () => {
```

test/court/court-draft.js
```diff
@@ -252,29 +252,45 @@ contract('Court', ([_, disputer, drafter, juror500, juror1000, juror1500, juror2
 
       const itHandlesDraftsProperlyForDifferentRequestedJurorsNumber = term => {
         context('when drafting all the requested jurors', () => {
-          const jurorsToBeDrafted = firstRoundJurorsNumber
-
           context('when drafting in one batch', () => {
-            itDraftsRequestedRoundInOneBatch(term, jurorsToBeDrafted)
+            const maxJurorsToBeDraftedPerBatch = firstRoundJurorsNumber
+
+            beforeEach('create court and dispute', async () => {
```
facuspagnuolo:
```suggestion
            beforeEach('set max number of jurors to be drafted per batch', async () => {
```

test/court/court-draft.js
```diff
@@ -283,29 +299,29 @@ contract('Court', ([_, disputer, drafter, juror500, juror1000, juror1500, juror2
...
           })
 
           itHandlesDraftsProperlyForDifferentRequestedJurorsNumber(term)
         })
 
         context('when the current term is after the randomness block number by 256 blocks', () => {
           beforeEach('move 256 blocks after the draft term', async () => {
-            await advanceBlocks(256)
+            // advance the block 1 less here because setting the maxJurorsToBeDraftedPerBatch will do the missing one
+            await advanceBlocks(255)
```
facuspagnuolo: I think we can simply comment that we are not advancing exactly X blocks cause there are some `before` that are actually increasing the block number. Note that we don’t have a reference to the max jurors per draft batch config here

contracts/court/Court.sol
bingen: Yes, good question. I was thinking about it for a while and decided to go this way because this feels to me almost like a constant, meaning that it should change very rarely. The only reasons I see for this to be changed are:
- Block gas limit changing. Actually it would only be needed if it goes down and below the gas needed for draft batches. For instance setting it to `100` jurors consumes ~2M gas and the way Court is configured now would actually cover the first 4 rounds using only one batch.
- Some weird EIPs changing gas prices, but again, it would be easy to set up a very safe value.
Besides (minor) right now Court config can be changed but it's kind of inconvenient as all config needs to provided.

test/court/court-draft.js
bingen: Sorry, I don't fully get it. The setting of max juror per draft batch is set in `itHandlesDraftsProperlyForDifferentRequestedJurorsNumber`, right after this `before`.

PR #182, opened 2019-10-23
Court: allow to have appeal collateral factors with decimals

### LINKED ISSUE #180: Court: Allow to use fractions for appeal factors
Currently, the court config only supports integer numbers for the appeal and appeal confirmation factors. We could change this into per mille values which will allow us to have the best of both worlds in case we want to have collaterals of less than 2x of the fees

bingen:
Fixes #180.


contracts/lib/PctHelpers.sol
```diff
@@ -16,6 +16,10 @@ library PctHelpers {
         return self.mul(uint256(_pct)) / PCT_BASE;
     }
 
+    function pct256(uint256 self, uint256 _pct) internal pure returns (uint256) {
```
facuspagnuolo: I think we can simply call this one `pct` and leverage Solidity overloading here. Additionally, we should reuse it in the `pct` method above (for `uint16`)

bingen: > I think we can simply call this one pct and leverage Solidity overloading here.
I was tempted, but not a big fan, mostly due to some problems I've had before with web3 and overloading. Anyway, I'll do it.
> Additionally, we should reuse it in the pct method above (for uint16)
Sure.

bingen: > I hope the compiler does a good job optimizing this

Not really: I measured an increase of 71 gas per call. Do you still want me to reuse it?

bingen: > I think we can simply call this one pct and leverage Solidity overloading here. 

It's giving me this error:

```
aragon-court/contracts/court/Court.sol:1191:28: TypeError: Member "pct" not unique after argument-dependent lookup in uint256.
            penalties[i] = minActiveBalance.pct(_penaltyPct).mul(_round.jurorsStates[juror].weight);
                           ^------------------^
```
So I'm going to revert it (unless there is any trick I don't know)

PR #188, opened 2019-10-25

bingen:
There's no real need for interval zero check, but the consistent
response to such a request is an empty array, as an interval of zero
length is empty.

To choose between this one, #189 or #190.


bingen:
Closed in favor of #190.

PR #189, opened 2019-10-25

bingen:
There's no need for interval zero check.

To choose between this one, #188 or #190


bingen:
Closed in favor of #190.

PR #190, opened 2019-10-25

bingen:
There's no real need for interval zero check, but if it was, the
search wouldn't make any sense.

To choose between this one, #188 or #189.


bingen:
For more context, see discussion [here](https://github.com/aragon/aragon-court/pull/176#discussion_r337392811)

PR #191, opened 2019-10-25
Court: Implement withdrawal lock for coherent jurors in final round

### LINKED ISSUE #165: Court: Implement withdrawal lock for coherent jurors in final round
In order to make 51% attacks more expensive, we can lock jurors that are coherent in the final round for a configurable number of terms. This is not an actual protection, since honest jurors will still be slashed, but it will allow honest jurors to exit the bonding curve before the attackers can withdraw from the Court.

In order to make things simpler, I think we can just prevent these jurors from withdrawing any juror tokens from the registry (regardless of the active stake when they voted in the final round)

bingen:
Fixes #165.


contracts/court/Court.sol
```diff
@@ -409,16 +409,24 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
...
+        _treasury().assign(config.fees.token, _juror, round.jurorFees.mul(jurorState.weight) / coherentJurors);
+
+        // Set the lock for final round
+        if (!_isRegularRound(_roundId, config)) {
+            // Round end term id (as it's final there's no draft delay nor appeal) plus the lock period
+            DisputesConfig memory disputesConfig = config.disputes;
+            uint64 finalRoundLockTermId = round.draftTermId +
+                disputesConfig.commitTerms + disputesConfig.revealTerms + disputesConfig.finalRoundLockTerms;
+            jurorsRegistry.finalRoundLock(_juror, finalRoundLockTermId);
+        }
```
facuspagnuolo: Shouldn't this logic be within `if (collectedTokens > 0)`?

contracts/registry/JurorsRegistry.sol
```diff
@@ -58,10 +59,11 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
...
     */
     struct Juror {
-        uint256 id;                 // key in the jurors tree used for drafting
-        uint256 lockedBalance;      // maximum amount of tokens that can be slashed based on the juror's drafts
-        uint256 availableBalance;   // available tokens that can be withdrawn at any time
+        uint256 id;                  // key in the jurors tree used for drafting
+        uint256 lockedBalance;       // maximum amount of tokens that can be slashed based on the juror's drafts
+        uint256 availableBalance;    // available tokens that can be withdrawn at any time
         DeactivationRequest deactivationRequest;
+        uint64 finalRoundLockTermId; // coherent jurors with final ruling in a final round will be locked for some time to prevent 51% attacks
```
facuspagnuolo: What do you think about renaming it to `withdrawalsLockTermId`?

contracts/registry/JurorsRegistry.sol
```diff
@@ -334,6 +336,16 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
         _setTotalActiveBalanceLimit(_totalActiveBalanceLimit);
     }
 
+    /**
+    * @notice Lock juror who voted in a final round and was coherent with the final ruling (to prevent 51% attacks)
+    * @param _juror Address of the juror to be locked
+    * @param _finalRoundLockTermId Term id which the juror will be locked until
+    */
+    function finalRoundLock(address _juror, uint64 _finalRoundLockTermId) external onlyCourt {
```
facuspagnuolo: I'd try to avoid the final round terminology here, we can simply talk about locking withdrawals:
```
/**
* @notice Lock `_juror`'s withdrawals until term #`_termId`
* @param _juror Address of the juror to be locked
* @param _termId Term id until which the juror's withdrawals will be locked
*/
function lockWithdrawals(address _juror, uint64 _termId) external onlyCourt {
```

contracts/registry/JurorsRegistry.sol
```diff
@@ -666,7 +678,14 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
         // the current term this time since deactivation requests always work with future terms, which means that if
         // the current term is outdated, it will never match the deactivation term id. We avoid ensuring the term here
         // to avoid forcing jurors to do that in order to withdraw their available balance.
-        _processDeactivationRequest(_juror, _getLastEnsuredTermId());
+        // Same applies to final round locks.
+        uint64 lastEnsuredTermId = _getLastEnsuredTermId();
+
+        // Check that juror is not locked due to a final round
```
facuspagnuolo:
```suggestion
        // Check that juror's withdrawals are not locked
```

contracts/registry/JurorsRegistry.sol
```diff
@@ -666,7 +678,14 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
         // the current term this time since deactivation requests always work with future terms, which means that if
         // the current term is outdated, it will never match the deactivation term id. We avoid ensuring the term here
         // to avoid forcing jurors to do that in order to withdraw their available balance.
-        _processDeactivationRequest(_juror, _getLastEnsuredTermId());
+        // Same applies to final round locks.
```
facuspagnuolo:
```suggestion
```

contracts/registry/JurorsRegistry.sol
```diff
@@ -666,7 +678,14 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
...
+        uint64 lastEnsuredTermId = _getLastEnsuredTermId();
+
+        // Check that juror is not locked due to a final round
+        uint64 finalRoundLockTermId = jurorsByAddress[_juror].finalRoundLockTermId;
+        require(finalRoundLockTermId == 0 || finalRoundLockTermId < lastEnsuredTermId, ERROR_FINAL_ROUND_LOCK);
```
facuspagnuolo: I think we can simply ask for `finalRoundLockTermId < lastEnsuredTermId` right? I mean, since deactivations are always processed for the next term, I think we can ensure this won't be executed unless the current term id is greater than 0

test/court/court-settle-round.js
```diff
@@ -314,6 +315,47 @@ contract('Court', ([_, disputer, drafter, appealMaker, appealTaker, juror500, ju
...
+                        await court.settleReward(disputeId, roundId, juror.address)
+                        await courtHelper.jurorsRegistry.deactivate(0, { from: juror.address }) // deactivate all
+                      }
+                    })
+
+                    it('locks only after final round lock period', async () => {
+                      // fails to withdraw on next term
+                      await courtHelper.passTerms(bn(1))
+                      for(const juror of expectedWinningJurors) {
+                        assertRevert(courtHelper.jurorsRegistry.unstake(amount, data, { from: juror.address }))
```
facuspagnuolo:
```suggestion
                        await assertRevert(courtHelper.jurorsRegistry.unstake(amount, data, { from: juror.address }))
```

test/court/court-settle-round.js
```diff
@@ -314,6 +315,47 @@ contract('Court', ([_, disputer, drafter, appealMaker, appealTaker, juror500, ju
...
+
+                      // fails to withdraw on last locked term
+                      const { draftTerm } = await court.getRound(disputeId, roundId)
+                      const lastLockedTermId = draftTerm
+                            .add(courtHelper.commitTerms)
+                            .add(courtHelper.revealTerms)
+                            .add(courtHelper.finalRoundLockTerms)
+                      await courtHelper.setTerm(lastLockedTermId)
+                      for(const juror of expectedWinningJurors) {
+                        assertRevert(courtHelper.jurorsRegistry.unstake(amount, data, { from: juror.address }))
```
facuspagnuolo:
```suggestion
                        await assertRevert(courtHelper.jurorsRegistry.unstake(amount, data, { from: juror.address }))
```

test/court/court-settle-round.js
facuspagnuolo: Let's add the error message here

test/court/court-settle-round.js
facuspagnuolo: Let's add the error message here

contracts/court/Court.sol
bingen: Not really. For the final round, if the juror can be rewarded, it means that he commited a vote, and therefore there are collected tokens. That condition only makes sense for the regular round.

contracts/registry/JurorsRegistry.sol
bingen: Nope, it was actually that way in the first version, but some tests were failing. Although it's quite en edge case, a juror could stake and unstake during term id 0 (like he changed his mind or whatever).

contracts/registry/JurorsRegistry.sol
bingen: Well, I don't agree that much. Those contracts are separated mostly because of gas issues, but are conceptually tightly coupled, it's not that this lock is going to be reused for other use cases, and I find it's good, when reading this code, to be aware what this is intended for.
But, anyway, changing it.

contracts/court/Court.sol
bingen: Btw, for that reason, I'm leaving references to final round in comments as we already have a bunch of them.

contracts/registry/JurorsRegistry.sol
bingen: Btw, for that reason, I'm leaving references to final round in comments as we already have a bunch of them.

contracts/court/Court.sol
bingen: This comment was placed in the wrong place, sorry, copied below to where it belongs.

PR #192, opened 2019-10-25
Registry: allow jurors in deactivation process to be drafted

### LINKED ISSUE #187: Jurors Registry: future deactivation vs present draft
I wonder what's exactly the point of deactivation in the future (next term). I thought the idea was to keep terms consistent so active tokens and therefore drafts probability don't change in the middle of a term.
But with this:
https://github.com/aragon/aragon-court/blob/development/contracts/registry/JurorsRegistry.sol#L727
if a juror sends a full deactivation request during term `t`, the juror won't be able to withdraw until term `t+1`, but won't be either be able to be drafted during term `t`.
I see 2 solutions here:

- Deactivate immediately. This would improve juror UX, as funds would be immediately available and deactivation and withdrawal could happen in the same transaction, but it would miss the aforementioned point of term integrity.
- Lowering next term deactivation request amount, instead of just checking it. I would go for this one as I think it was the initial intention.

bingen:
Fix bug in drafting with deactivation checks. Before a juror with a
pending deactivation request big enough wouldn't be able to be drafted.

Fixes #187.


contracts/registry/JurorsRegistry.sol
```diff
@@ -729,7 +729,14 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
                 uint256 nextTermDeactivationRequestAmount = _deactivationRequestedAmountForTerm(juror, _draftParams.termId + 1);
 
                 // Check if juror has enough active tokens to lock the requested amount for the draft, skip it otherwise.
-                if (activeBalances[i].sub(nextTermDeactivationRequestAmount) >= newLockedBalance) {
+                // TODO: stack too deep
+                //uint256 currentActiveBalance = activeBalances[i];
+                if (activeBalances[i] >= newLockedBalance) {
+                    // If this reverts there's a bug somewhere, but let's keep SafeMath for extra safety
```
facuspagnuolo: I think we can avoid saying that about safety, I'd add a comment clarifying
```suggestion
                    // Check if the amount of active tokens for the next term is enough to lock the required amount for 
                    // the draft. Otherwise, reduce the requested deactivation amount of the next term.
```

test/registry/jurors-registry-drafting.js
```diff
@@ -478,63 +478,63 @@ contract('JurorsRegistry', ([_, juror500, juror1000, juror1500, juror2000, juror
...
                       const previousSelectedJurors = 3
 
-                      itReturnsExpectedJurors({ previousSelectedJurors, batchRequestedJurors, roundRequestedJurors })
+                      itReturnsExpectedJurors({ disputeId, previousSelectedJurors, batchRequestedJurors, roundRequestedJurors })
                     })
-                  })
 
-                  context('when some jurors do not have been enough balance to be drafted again', () => {
-                    context('when it is due to other drafts', () => {
+                    context('when some jurors have deactivation requests', () => {
```
facuspagnuolo: We can wrap the two contexts above into a `context('when jurors don't have deactivation requests', () => {` to make it symmetrical with this one

test/registry/jurors-registry-drafting.js
```diff
@@ -478,63 +478,63 @@ contract('JurorsRegistry', ([_, juror500, juror1000, juror1500, juror2000, juror
...
 
-                        itReturnsExpectedJurors({ previousSelectedJurors, batchRequestedJurors, roundRequestedJurors })
+                        itReturnsExpectedJurors({ disputeId, previousSelectedJurors, batchRequestedJurors, roundRequestedJurors })
                       })
                     })
+                  })
 
-                    context('when it is due to deactivation requests', () => {
+                  context('when some jurors do not have been enough balance to be drafted again', () => {
+                    context('when it is due to other drafts', () => {
```
facuspagnuolo: I think we can now drop this `context` block

test/registry/jurors-registry-staking.js
```diff
@@ -1088,6 +1094,7 @@ contract('JurorsRegistry', ([_, juror, anotherJuror]) => {
 
           await registry.unstake(amount, data, { from })
 
+          await controller.mockIncreaseTerm()
```
facuspagnuolo: I don't think we need to increase a term for `unstake`

test/registry/jurors-registry-slashing.js
```diff
@@ -236,10 +236,18 @@ contract('JurorsRegistry', ([_, juror, secondJuror, thirdJuror, anyone]) => {
...
         it('decreases the unlocked balance of the juror', async () => {
-          const previousUnlockedActiveBalance = await registry.unlockedActiveBalanceOf(juror)
+          const pendingDeactivation = await registry.getDeactivationRequest(juror)
+          const currentTermId = await controller.getLastEnsuredTermId()
+
+          let pendingDeactivationAmount = bn(0)
+          if (pendingDeactivation.availableTermId.gt(currentTermId)) {
+            pendingDeactivationAmount = pendingDeactivation.amount
+          }
+          const previousUnlockedActiveBalance = (await registry.unlockedActiveBalanceOf(juror)).sub(pendingDeactivationAmount)
```
facuspagnuolo: Let's add a comment here saying that we are deducting a potential deactivation request because `unlockedActiveBalanceOf` will always return the very last state of the juror's balances

test/registry/jurors-registry-drafting.js
bingen: Sorry, which ones?

test/registry/jurors-registry-slashing.js
bingen: Well, it's actually the opposite, but I'll add the comment.

Issue #185, opened 2019-10-24
Config: Optimize getters for the Court


facuspagnuolo:
Currently, we are exposing one single getter retrieving the whole config every time. We could optimize this having different getters for each particular consumer

PR #195, opened 2019-10-26
Court: move minActiveBalance from JurorsRegistry to Config

### LINKED ISSUE #183: Registry: Checkpoint min active balance
Given that we're allowing to change the min active balance in #132, we should checkpoint with the Court terms to make sure old disputes are not accessing to changing value

bingen:
Fixes #183.

### DIFF
--- a/contracts/controller/clock/CourtClock.sol
+++ b/contracts/controller/clock/CourtClock.sol
@@ -52,10 +52,14 @@ contract CourtClock is IClock, TimeHelpers {
 
     /**
     * @dev Constructor function
-    * @param _termDuration Duration in seconds per term
-    * @param _firstTermStartTime Timestamp in seconds when the court will open (to give time for juror on-boarding)
+    * @param _termParams Array containing:
+    *        0. _termDuration Duration in seconds per term
+    *        1. _firstTermStartTime Timestamp in seconds when the court will open (to give time for juror on-boarding)
     */
-    constructor(uint64 _termDuration, uint64 _firstTermStartTime) public {
+    constructor(uint64[2] memory _termParams) public {
+        uint64 _termDuration = _termParams[0];
+        uint64 _firstTermStartTime = _termParams[1];
+
         require(_termDuration < MAX_TERM_DURATION, ERROR_TERM_DURATION_TOO_LONG);
         require(_firstTermStartTime >= getTimestamp64() + _termDuration, ERROR_BAD_FIRST_TERM_START_TIME);
         require(_firstTermStartTime <= getTimestamp64() + MAX_FIRST_TERM_DELAY_PERIOD, ERROR_BAD_FIRST_TERM_START_TIME);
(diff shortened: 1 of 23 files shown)


contracts/controller/config/CourtConfig.sol
```diff
@@ -162,6 +177,9 @@ contract CourtConfig is IConfig, CourtConfigData {
             require(_roundStateDurations[i] > 0 && _roundStateDurations[i] < MAX_ADJ_STATE_DURATION, ERROR_LARGE_ROUND_PHASE_DURATION);
         }
 
+        // Make sure min active balance is not zero
+        require(_minActiveBalance > 0, ERROR_ZERO_MIN_ACTIVE_BALANCE);
```
facuspagnuolo: What do you think about allowing it to be zero? If it's not harmful I'd allow it

contracts/controller/clock/CourtClock.sol
```diff
@@ -52,10 +52,14 @@ contract CourtClock is IClock, TimeHelpers {
...
-    * @param _termDuration Duration in seconds per term
-    * @param _firstTermStartTime Timestamp in seconds when the court will open (to give time for juror on-boarding)
+    * @param _termParams Array containing:
+    *        0. _termDuration Duration in seconds per term
+    *        1. _firstTermStartTime Timestamp in seconds when the court will open (to give time for juror on-boarding)
     */
-    constructor(uint64 _termDuration, uint64 _firstTermStartTime) public {
+    constructor(uint64[2] memory _termParams) public {
+        uint64 _termDuration = _termParams[0];
+        uint64 _firstTermStartTime = _termParams[1];
```
facuspagnuolo: Nit: let's avoid the `_` prefix here

bingen:
Some gas increase:

| Function            | Without heartbeat | With heartbeat |
|---------------------|-------------------|---|
|       createDispute |           216,311 | 272,484 |
|               draft |           344,858 |
|              commit |            92,881 | 149,235 |
|              reveal |           105,614 | 161,968 |
|        createAppeal |            88,534 | 144,888 |
|       confirmAppeal |           193,371 | 249,725 |
|       executeRuling |            67,166 | 123,520 |
|     settlePenalties |           211,758 | 261,360 |
|        settleReward |            87,265 | 87,265 |
| settleAppealDeposit |            81,253 | 80,989 |

:'(

contracts/controller/config/CourtConfigData.sol
```diff
@@ -29,5 +29,6 @@ contract CourtConfigData {
         uint256 maxRegularAppealRounds;         // Before the final appeal
         uint256 appealCollateralFactor;         // Permyriad multiple of juror fees required to appeal a preliminary ruling (‱ - 1/10,000)
         uint256 appealConfirmCollateralFactor;  // Permyriad multiple of juror fees required to confirm appeal (‱ - 1/10,000)
+        uint256 minActiveBalance;               // Minimum amount of tokens jurors have to activate to participate in the Court
```
facuspagnuolo: I think I'd place it at the `Config` level instead, note that it will simplify things on the `JurorsRegistry`'s end since you don't need to fetch the whole config, but the `Court` won't change

contracts/controller/config/CourtConfig.sol
bingen: We are dividing by it when computing weight for final rounds.

facuspagnuolo: True, let's keep it then 😅

contracts/controller/config/CourtConfigData.sol
bingen: Sounds good. Let's do it in #185 then.

contracts/controller/clock/CourtClock.sol
bingen: `termDuration` is already defined as a global variable, and those 2 are definitely playing the role of parameters. I just used them for better readability.

PR #198, opened 2019-10-26
Contracts: Stack too deep issues

### LINKED ISSUE #43: Contracts: Stack too deep issues

facuspagnuolo:
Fixes #43 

This one was the last one, at least commented in the codebase. I suggest closing the issue after merging this PR, if we find more instances for this problem in the future we can open a new one. Otherwise, it will be a bit cumbersome tracking them.


facuspagnuolo:
The "stack too deep issue" comes from not being allowed to map a certain amount of variables into the stack, but since the `DraftParams` object is already in memory, we don't need to cache `requestedJurors` again. 

Apart from that, I checked the gas costs of drafting using the deterministic version of the [gas tests](https://github.com/aragon/aragon-court/pull/199/commits/6d5a151e821dc0a80b77a74bf592a1d047b8e16d) and they were decreased by 110 gas.

Then, I'd say it is fixed 🙃

bingen:
> but since the `DraftParams` object is already in memory, we don't need to cache `requestedJurors` again.

Ok, but `activeBalances` was also in memory, so what's the difference? Does it make any difference accessing an array element or a struct element?
It seems to me that if we could cache both that would be the most gas efficient, that's why I think it's not actually fixed, although certainly not being `SLOAD`s makes it not that big of a deal.

PR #199, opened 2019-10-27
Config: Optimize min active balance getter

facuspagnuolo:
First intent towards #185

### DIFF
--- a/contracts/controller/Controlled.sol
+++ b/contracts/controller/Controlled.sol
@@ -183,10 +183,20 @@ contract Controlled is IsContract, CourtConfigData {
             maxRegularAppealRounds: _roundParams[2],
             finalRoundLockTerms: _roundParams[3],
             appealCollateralFactor: _appealCollateralParams[0],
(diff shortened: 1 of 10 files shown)


facuspagnuolo:
@bingen yea, there was a problem in particular with those two functions since their costs are proportional to the draft output, which wasn't being deterministic. 

What I did was to force a certain randomness seed to make sure the draft returns always the same jurors (see [here](https://github.com/aragon/aragon-court/pull/199/commits/6d5a151e821dc0a80b77a74bf592a1d047b8e16d)), and evaluate how the gas cost changes after that with the custom config getter (see [here](https://github.com/aragon/aragon-court/pull/199/commits/1815a706ef6f3ad005c94113a6d6b886aa1cd893#diff-7)). As you can see the drafting is improved by 10k, which is aligned to what we were expecting.

facuspagnuolo:
@bingen interesting, honestly I don't. However, I think we are good to merge since we are lowering that one in https://github.com/aragon/aragon-court/pull/205

PR #204, opened 2019-10-31
Chore: Drop aragonOS dependency

facuspagnuolo:
Following up a conversation I had with @bingen and @sohkai, we are not planning to release a beta version nor release-candidate of aragonOS soon. Therefore, dropping this dependency and moving the libraries we are reusing manually.


contracts/lib/os/SafeERC20.sol
```diff
@@ -0,0 +1,156 @@
...
+                }
+
+                // Not sure what was returned: don't mark as success
+                default { }
+            }
+        }
+        return ret;
+    }
+
+    function staticInvoke(address _addr, bytes memory _calldata)
```
sohkai: Note that this is pretty useless in 0.5.0 because `view` or `pure` calls are automatically made statically.

It may still be a slight gas optimization, but I'd be in favour of removing all related code to this.

package.json
```diff
@@ -15,11 +15,10 @@
   "pre-push": [
     "lint"
   ],
-  "dependencies": {
-    "@aragon/os": "aragon/aragonOS#alpha_upgrade_solidity_5"
-  },
+  "dependencies": {},
```
sohkai: 🙌

Issue #107, opened 2019-09-01
Court: Optimize storage loadings


facuspagnuolo:
`SLOAD` operations costs are going to be increased after Istambul due to EIP-1884. Bearing in mind that we had to repeat multiple loadings due to stack-to-deep issues, we should consider strongly using ABI Encoder V2, or any other workaround to these issues. Besides, it would be good to review all the contracts and make sure storage loadings are optimized as much as possible.

facuspagnuolo:
Closing this issue since the current implementation had a lot of iterations over this

PR #205, opened 2019-10-31
Court: Optimize config getters

### LINKED ISSUE #185: Config: Optimize getters for the Court

facuspagnuolo:
Fixes #185 

This PR includes the following changes:
- Optimized config getter for creating disputes
- Optimized config getter for drafting
- Make sure config is fetched only once in the rest of the Court entry points (a step towards #107)

### DIFF
--- a/contracts/court/Court.sol
+++ b/contracts/court/Court.sol
@@ -203,8 +203,8 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
         Dispute storage dispute = disputes[disputeId];
         dispute.subject = _subject;
         dispute.possibleRulings = _possibleRulings;
-        Config memory config = _getConfigAt(draftTermId);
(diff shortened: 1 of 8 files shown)


contracts/court/Court.sol
```diff
@@ -657,9 +666,14 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
             uint256 confirmAppealDeposit
         )
     {
+        _checkRoundExists(_disputeId, _roundId);
```
bingen: :heart: :raised_hands:  (why this change now though?)

bingen: Also, why only here?

facuspagnuolo: stack too deep my friend 😓

PR #211, opened 2019-11-10
Clock: Allow to delay the court start time

### LINKED ISSUE #209: Clock: Allow to change the first term start date if it wasn't reached
Simply allow the config governor to modify this date in the future in case it wasn't reached

facuspagnuolo:
Fixes #209


contracts/controller/Controller.sol
```diff
@@ -176,6 +176,14 @@ contract Controller is IsContract, CourtClock, CourtConfig {
         );
     }
 
+    /**
+    * @notice Delay the Court start time to `_newFirstTermStartTime`
+    * @param _newFirstTermStartTime New timestamp in seconds when the court will open
+    */
+    function delayStartTime(uint64 _newFirstTermStartTime) external onlyConfigGovernor {
+        _delayStartTime(_newFirstTermStartTime);
```
bingen: We are not reusing `_delayStartTime`. Why do we need the internal function?

facuspagnuolo: This is due to the architecture model we have, the internal function `_delayStartTime` is defined by the `CourtClock`, while the public is defined by the `Controller` where the authorization modifier is being added

bingen: Oh, sure, sorry I missed that detail!

PR #214, opened 2019-11-11
Court: Defined Arbitrable interface

facuspagnuolo:
*This is the second of a 4-series PRs, following https://github.com/aragon/aragon-court/pull/213*

This PR aims to simplify and define the `IArbitrable` interface we are using for the Court while solving the pending `TODO`s about its validations


contracts/court/Court.sol
```diff
@@ -546,15 +548,14 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
 
     /**
     * @dev Tell the amount of token fees required to create a dispute
-    * @param _draftTermId Term ID in which the dispute will be drafted
     * @return feeToken ERC20 token used for the fees
-    * @return jurorFees Total amount of fees to be distributed between the winning jurors of a round
     * @return totalFees Total amount of fees for a regular round at the given term
     */
-    function getDisputeFees(uint64 _draftTermId) external view returns (ERC20 feeToken, uint256 jurorFees, uint256 totalFees) {
```
bingen: Why are we changing this?

contracts/court/Court.sol
```diff
@@ -182,31 +186,29 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
...
-    * @notice Create a dispute over `_subject` with `_possibleRulings` possible rulings in next term
-    * @dev Create a dispute to be drafted in a future term
-    * @param _subject Arbitrable subject being disputed
+    * @notice Create a dispute over the Arbitrable sender with `_possibleRulings` possible rulings for the next term
     * @param _possibleRulings Number of possible rulings allowed for the drafted jurors to vote on the dispute
     * @param _metadata Optional metadata that can be used to provide additional information on the dispute to be created
     * @return Dispute identification number
     */
-    function createDispute(IArbitrable _subject, uint8 _possibleRulings, bytes calldata _metadata) external returns (uint256) {
-        // TODO: Limit the min amount of terms before drafting (to allow for evidence submission)
```
bingen: Is this fixed? Did we dismissed it?

test/court/court-gas.js
```diff
@@ -53,7 +54,7 @@ contract('Court', ([_, sender, disputer, drafter, appealMaker, appealTaker, juro
           assertBn(neededTransitions, 0, 'needed transitions does not match')
         })
 
-        itCostsAtMost('createDispute', 210e3, () => court.createDispute(arbitrable.address, 2, '0x', { from: sender }))
+        itCostsAtMost('createDispute', 251e3, () => arbitrable.createDispute(2, '0x', { from: sender }))
```
bingen: Why are we having this big increase? Is it because we are using another contract in between?

facuspagnuolo: yup, actually two more contracts, arbitrable => oracle => disputes manager

contracts/court/Court.sol
facuspagnuolo: I was thinking of moving the idea of scheduling drafts for the next term always to have a configuration for it... otherwise, I don't see a validation that makes sense here

contracts/court/Court.sol
facuspagnuolo: Cause it wasn't being used, as a consumer I want to know how many fees I need to pay in total only

PR #215, opened 2019-11-11
Court: Implement disputes resolution oracle interface

facuspagnuolo:
*This is the third of a 4-series PRs, following https://github.com/aragon/aragon-court/pull/214*

This PR aims to provide support for the `IDisputeResolutionProtocol` required by the `ProposalAgreement` spec, implemented at a `Controller` level. Note that some changes were needed to be allowed to guarantee a single entry-point for users.

### DIFF
--- a/contracts/controller/oracle/DisputeResolutionOracle.sol
+++ b/contracts/controller/oracle/DisputeResolutionOracle.sol
@@ -1,0 +1,82 @@
+pragma solidity ^0.5.8;
+
+import "./IDisputeResolutionOracle.sol";
+import "../../court/ICourt.sol";
+import "../../arbitration/IArbitrable.sol";
+import "../../subscriptions/ISubscriptions.sol";
+import "../../lib/os/Uint256Helpers.sol";
+
+
+contract DisputeResolutionOracle is IDisputeResolutionOracle {
+    using Uint256Helpers for uint256;
+
+    string private constant ERROR_SENDER_NOT_ARBITRABLE = "CT_SENDER_NOT_ARBITRABLE";
+    string private constant ERROR_SUBSCRIPTION_NOT_PAID = "CT_SUBSCRIPTION_NOT_PAID";
+
+    // Arbitrable interface ID based on ERC-165
+    bytes4 private constant ARBITRABLE_INTERFACE_ID = bytes4(0x311a6c56);
+
+    /**
+    * @dev Create a dispute over the Arbitrable sender with a number of possible rulings
+    * @param _possibleRulings Number of possible rulings allowed for the drafted jurors to vote on the dispute
+    * @param _metadata Optional metadata that can be used to provide additional information on the dispute to be created
+    * @return Dispute identification number
+    */
+    function createDispute(uint256 _possibleRulings, bytes calldata _metadata) external returns (uint256) {
+        IArbitrable subject = IArbitrable(msg.sender);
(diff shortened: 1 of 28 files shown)


contracts/controller/oracle/DisputeResolutionOracle.sol
```diff
@@ -0,0 +1,82 @@
+pragma solidity ^0.5.8;
+
+import "./IDisputeResolutionOracle.sol";
+import "../../court/ICourt.sol";
+import "../../arbitration/IArbitrable.sol";
+import "../../subscriptions/ISubscriptions.sol";
+import "../../lib/os/Uint256Helpers.sol";
+
+
+contract DisputeResolutionOracle is IDisputeResolutionOracle {
```
bingen: So this is the old `IArbitrator`, right? I think I liked better that other name, simpler and matches better with its pair `IArbitrable`. But as you want.

contracts/court/Court.sol
```diff
@@ -186,15 +182,13 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
...
+    * @param _subject Arbitrable instance creating the dispute
     * @param _possibleRulings Number of possible rulings allowed for the drafted jurors to vote on the dispute
     * @param _metadata Optional metadata that can be used to provide additional information on the dispute to be created
     * @return Dispute identification number
     */
-    function createDispute(uint8 _possibleRulings, bytes calldata _metadata) external returns (uint256) {
-        IArbitrable subject = IArbitrable(msg.sender);
-        require(subject.supportsInterface(ARBITRABLE_INTERFACE_ID), ERROR_SENDER_NOT_ARBITRABLE);
-
+    function createDispute(IArbitrable _subject, uint8 _possibleRulings, bytes calldata _metadata) external onlyController returns (uint256) {
```
bingen: I would add a comment somewhere saying that this works under the assumption that `_subject` is correct: it's Arbitrable and it's the original `msg.sender`

contracts/court/Court.sol
```diff
@@ -917,14 +910,15 @@ contract Court is ControlledRecoverable, ICRVotingOwner {
     }
 
     /**
-    * @dev Internal function to execute a deposit of tokens from the msg.sender to the Court treasury contract
+    * @dev Internal function to execute a deposit of tokens from an account to the Court treasury contract
+    * @param _from Address transferring the amount of tokens
     * @param _token ERC20 token to execute a transfer from
-    * @param _amount Amount of tokens to be transferred from the msg.sender to the Court treasury
+    * @param _amount Amount of tokens to be transferred from the sender to the Court treasury
```
bingen: Well, now it's not the sender, right? It's the `_from`, which in one of the origins is `subject`.

contracts/court/Court.sol
facuspagnuolo: But the `DisputesManager` doesn't need to ensure that, it is actually validated in the Oracle

contracts/controller/oracle/DisputeResolutionOracle.sol
facuspagnuolo: We can keep it but it's completely different from the original one

facuspagnuolo: I'm renaming it in the next PR to avoid having tons of conflict

PR #217, opened 2019-11-12
Chore: remove Kleros references

izqui:


readme.md
```diff
@@ -25,9 +25,6 @@ A design goal of the mechanism is to require very few jurors to adjudicate a dis
 
 Even though Aragon Court could theoretically resolve any type of binary dispute, in its first deployments it will be used to arbitrate **Proposal Agreements.** These agreements require entities creating a proposal in an organization to agree to its specific rules around proposal creation, putting some collateral at stake that could be lost if the Court finds the proposal invalid.
 
-This first version of Aragon Court was inspired by [Kleros protocol](https://github.com/kleros/kleros), although major changes to the mechanism were made to increase some important security aspects.
```
sohkai: We may reword this, but I'd leave the reference. Give credit where credit is due; we would not be where we are today without their implementation.

izqui: We have given credit publicly repeatedly, I just don't think we have to keep doing so in the repo.

izqui: @facuspagnuolo thoughts?

facuspagnuolo: I think it's okay based on the repeated times we mentioned their work

facuspagnuolo: I think it's okay based on the repeated times we mentioned their work

PR #220, opened 2019-11-15
Disputes: Implement evidence period

facuspagnuolo:
This PR aims to implement the evidence submission period for disputes right before the first draft


contracts/arbitration/IArbitrable.sol
```diff
@@ -1,8 +1,21 @@
...
 
+import "./IArbitrator.sol";
+
 
 interface IArbitrable {
-    // bytes4 constant INTERFACE_ID = 0x311a6c56;
+    // bytes4 constant INTERFACE_ID = 0x88f3ee69;
+
+    event Ruled(IArbitrator indexed arbitrator, uint256 indexed disputeId, uint256 ruling);
+    event EvidenceSubmitted(uint256 indexed disputeId, bytes evidence, bool finished);
```
bingen: Can we add a reference to who submits the evidence?

contracts/court/AragonCourt.sol
```diff
@@ -97,7 +99,18 @@ contract AragonCourt is Controller, IArbitrator {
...
     /**
-    * @dev Execute the arbitrable associated to a dispute based on its final ruling
+    * @notice Close the evidence period of dispute #`_disputeId`
+    * @param _disputeId Identification number of the dispute to close its evidence submitting period
+    */
+    function closeEvidencePeriod(uint256 _disputeId) external {
+        IDisputeManager disputeManager = IDisputeManager(_getDisputeManager());
+        (IArbitrable subject,,,,,) = disputeManager.getDispute(_disputeId);
+        require(subject == IArbitrable(msg.sender), ERROR_SENDER_NOT_DISPUTE_SUBJECT);
+        disputeManager.closeEvidencePeriod(_disputeId);
```
bingen: Maybe we can send `msg.sender` to `disputeManager.closeEvidencePeriod` and do the check inside. This way we save 2 calls to `disputeManager`.

contracts/court/config/CourtConfig.sol
```diff
@@ -164,17 +166,19 @@ contract CourtConfig is IConfig, CourtConfigData {
...
         internal
     {
-        // If the current term is not zero, changes must be scheduled at least 2 terms in the future.
-        // This way we can ensure that disputes scheduled for the next term won't have their config changed.
-        require(_currentTermId == 0 || _fromTermId > _currentTermId + 1, ERROR_TOO_OLD_TERM);
+        // If the current term is not zero, changes must be scheduled at least after the next evidence period.
+        // This way we can ensure that disputes already scheduled won't have their config changed.
+        uint64 evidenceTerms = configs[configIdByTerm[_currentTermId]].disputes.evidenceTerms;
+        // No need for SafeMath: round state durations are safely capped
+        require(_currentTermId == 0 || _fromTermId > _currentTermId + evidenceTerms, ERROR_TOO_OLD_TERM);
```
bingen: If we are going to use creation term for dispute configs we don't need this check.

contracts/disputes/DisputeManager.sol
```diff
@@ -181,26 +169,27 @@ contract DisputeManager is ControlledRecoverable, ICRVotingOwner, IDisputeManage
...
+    * @notice Create a dispute over `_subject` with `_possibleRulings` possible rulings
     * @param _subject Arbitrable instance creating the dispute
     * @param _possibleRulings Number of possible rulings allowed for the drafted jurors to vote on the dispute
     * @param _metadata Optional metadata that can be used to provide additional information on the dispute to be created
     * @return Dispute identification number
     */
     function createDispute(IArbitrable _subject, uint8 _possibleRulings, bytes calldata _metadata) external onlyController returns (uint256) {
         uint64 termId = _ensureCurrentTerm();
-        ISubscriptions subscriptions = _subscriptions();
-        require(subscriptions.isUpToDate(address(msg.sender)), ERROR_SUBSCRIPTION_NOT_PAID);
```
bingen: Why are we removing this?

contracts/disputes/DisputeManager.sol
```diff
@@ -212,6 +201,26 @@ contract DisputeManager is ControlledRecoverable, ICRVotingOwner, IDisputeManage
         return disputeId;
     }
 
+    /**
+    * @notice Close the evidence period of dispute #`_disputeId`
+    * @param _disputeId Identification number of the dispute to close its evidence submitting period
+    */
+    function closeEvidencePeriod(uint256 _disputeId) external onlyController {
+        uint256 roundId = 0;
+        _checkRoundExists(_disputeId, roundId);
```
bingen: You know I like it better here, but why isn't this as a modifier?

contracts/arbitration/IArbitrable.sol
facuspagnuolo: Sure!

contracts/court/AragonCourt.sol
facuspagnuolo: Yes, I did like you say at first, but then rolled back to this way due to size limits. Will try again.

contracts/court/config/CourtConfig.sol
facuspagnuolo: True, good catch!

contracts/disputes/DisputeManager.sol
facuspagnuolo: Cause it was already validated in the entry point (`AragonCourt`) :)

contracts/disputes/DisputeManager.sol
facuspagnuolo: I know, simply to reuse the `roundId` variable, it is redundant tho

PR #225, opened 2019-11-17
Disputes: Fix dispute manager errors prefix

### LINKED ISSUE #223: Prefix DisputeManager errors with 'DM' instead of 'CT'

facuspagnuolo:
Fixes #223


contracts/court/AragonCourt.sol
```diff
@@ -103,10 +102,9 @@ contract AragonCourt is Controller, IArbitrator {
     * @param _disputeId Identification number of the dispute to close its evidence submitting period
     */
     function closeEvidencePeriod(uint256 _disputeId) external {
+        IArbitrable subject = IArbitrable(msg.sender);
```
bingen: I think I would move this to the DisputeManager function.

PR #233, opened 2019-11-19
Audit: Apply code recommendations

### LINKED ISSUE #230: Audit: Code recommendations
### Code recommendations 
(report [reference](https://docs.google.com/document/d/1c7rEDCzJLuaxq5FHRRxYZ09rqityk9kcOOyPyfKLgbI/edit#heading=h.gw1chg2vi6o))

- [x] Rename variables which shadow function names (4d42357)
- [x] Use SafeMath wherever possible (1f84d8e and https://github.com/aragon/aragon-court/pull/233/commits/3cf17f8e1c8d4cee93e8ddc3d16aec818e381089)
- [x] Fix inconsistent function naming (8aab013)
- [x] Inline over-abstracted helper functions (a0ba4cf and 0368d49)
- [x] Rename `encryptVote` since it does not perform encryption (0368d49)
- [x] Update documentation for appeal collateral factors (c56c0ef)
- [x] Improve clock callback pattern (f2685a2)
- [x] CourtConfig does not implement IConfig (1ad8d60)
- [x] Fetch blocknumber only once when transitioning multiple terms (067471e)
- [x] Clearly document the pass by reference when drafting (c65ca3a)
- [x] Remove unused variables (b2925d2)
- [x] Use 32-bytes mask to decode vote IDs (2c0b0a8)
- [x] Fix voting module outcomes validation limit (9ef5f7a)
- [x] Consider enriching inline documentation (f52db4e)

facuspagnuolo:
Fixes #230


contracts/court/clock/CourtClock.sol
```diff
@@ -196,6 +198,7 @@ contract CourtClock is IClock, TimeHelpers {
         uint256 transitions = uint256(_maxRequestedTransitions < neededTransitions ? _maxRequestedTransitions : neededTransitions);
         require(transitions > 0, ERROR_INVALID_TRANSITION_TERMS);
 
+        uint64 blockNumber = getBlockNumber64();
```
bingen: What's the point of this?

contracts/lib/JurorsTreeSortition.sol
```diff
@@ -78,15 +78,10 @@ library JurorsTreeSortition {
         returns (uint256 low, uint256 high)
     {
         uint256 totalActiveBalance = tree.getRecentTotalAt(_termId);
-        // No need for SafeMath: the number of round requested jurors is always ensured to be greater than zero in the Court config
-        low = _selectedJurors.mul(totalActiveBalance) / _roundRequestedJurors;
+        low = _selectedJurors.mul(totalActiveBalance).div(_roundRequestedJurors);
```
bingen: Division by zero will already revert, right?

contracts/registry/JurorsRegistry.sol
```diff
@@ -688,11 +675,12 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
...
     * @dev Internal function to draft a set of jurors based on a given set of params
     * @param _params Params to be used for the jurors draft
-    * @param _jurors List of unique jurors selected for the draft
-    * @return Number of unique jurors selected for the draft. Note that this value may differ from the number of requested jurors
+    * @return length Number of unique jurors selected for the draft. Note that this value may differ from the number of requested jurors
+    * @return jurors List of jurors addresses selected for the draft
     */
-    function _draft(DraftParams memory _params, address[] memory _jurors) internal returns (uint256) {
-        uint256 length = 0;
+    function _draft(DraftParams memory _params) internal returns (uint256 length, address[] memory jurors) {
```
bingen: I think I have said this before, but I would just get rid of this function (as we have done with some others in this same PR) and put everything in the main `draft` function to avoid the overhead of passing and returning data.

test/court/court-gas.js
```diff
@@ -63,7 +63,7 @@ contract('AragonCourt', ([_, sender, drafter, appealMaker, appealTaker, juror500
           assertBn(neededTransitions, 1, 'needed transitions does not match')
         })
 
-        itCostsAtMost('createDispute', 308e3, () => arbitrable.createDispute(2, '0x', { from: sender }))
+        itCostsAtMost('createDispute', 274e3, () => arbitrable.createDispute(2, '0x', { from: sender }))
```
bingen: Why is this decreasing? I mean, adding safe math everywhere should actually increase gas, right? I see some others below decreasing too.

truffle-config.js
```diff
@@ -1,5 +1,6 @@
 const TruffleConfig = require('@aragon/truffle-config-v5/truffle-config')
 
 TruffleConfig.compilers.solc.version = '0.5.8'
+TruffleConfig.compilers.solc.settings.optimizer.runs = 5000 // DisputesManager module is hitting size limit with 10k runs
```
bingen: Oh, I thought we were still at `1`. So we were using the default `10k` from aragonOS? Why did we have to decrease it?This makes even more surprising that some gas measures above decreased.

contracts/lib/JurorsTreeSortition.sol
facuspagnuolo: yes but with an invalid opcode instead

facuspagnuolo: Yes but it will result in an invalid opcode instead

contracts/registry/JurorsRegistry.sol
facuspagnuolo: I tried but you it results in a stack too deep issue

test/court/court-gas.js
facuspagnuolo: Not sure if these diffs were left from before or if was due to the optimizer runs 🤷‍♂️

contracts/court/clock/CourtClock.sol
facuspagnuolo: Simply querying the block number outside the loop

truffle-config.js
facuspagnuolo: > So we were using the default 10k from aragonOS?

Yes

> Why did we have to decrease it?

It was because for the comment I added there: `DisputesManager module is hitting size limit with 10k runs`

contracts/court/clock/CourtClock.sol
bingen: Yes, cool, I overlooked that, thanks!

contracts/lib/JurorsTreeSortition.sol
bingen: Yes, I know, but security-wise (the purpose of this change) it's the same, right? I don't know how much gas difference there is due to this change though.
Also, we are not checking anything for mod division, so I don't see why we should do it for regular division.

truffle-config.js
bingen: > It was because for the comment I added there: DisputesManager module is hitting size limit with 10k runs

Haha, yes I read that comment, but I was wondering what in particular was the reason for the increase. Now I know that is because of adding SafeMath.

contracts/lib/JurorsTreeSortition.sol
facuspagnuolo: Yes, it's the same. True about `mod` we could implement something to support it for `SafeMath` if you want. In this case, I was mostly driven based on the definition we set in order to decide when to use or not SafeMath, only when we are checking it in the context of the function, since we are not checking `_roundRequestedJurors` here, I added it

bingen: > In this case, I was mostly driven based on the definition we set in order to decide when to use or not SafeMath, only when we are checking it in the context of the function, since we are not checking _roundRequestedJurors here, I added it

Yes, but the definition was for security issues, while here we are not solving any security issue.

contracts/registry/JurorsRegistry.sol
bingen: I just tried and it worked for me:
https://github.com/aragon/aragon-court/tree/remove_internal_draft

PR #244, opened 2019-12-07
Contracts: Improve emitted events

facuspagnuolo:
After working on the court subgraph I noticed there are some events we are triggering that are a bit loose and some others that we simply forgot to include certain data 🤷‍♂ 

I needed to adjust the `DisputesManager` and the `JurorsRegistry` only, I recommend reviewing by commit tho.

### DIFF
--- a/contracts/disputes/DisputeManager.sol
+++ b/contracts/disputes/DisputeManager.sol
@@ -106,6 +106,14 @@ contract DisputeManager is ControlledRecoverable, ICRVotingOwner, IDisputeManage
         bool settled;                  // Whether or not an appeal has been settled
     }
 
+    struct DraftParams {
+        uint256 disputeId;            // Identification number of the dispute to be drafted
+        uint256 roundId;              // Identification number of the round to be drafted
+        uint64 termId;                // Identification number of the current term of the Court
+        bytes32 draftTermRandomness;  // Randomness of the term in which the dispute was requested to be drafted
+        DraftConfig config;           // Draft config of the Court at the draft term
+    }
+
     struct NextRoundDetails {
         uint64 startTerm;              // Term ID from which the next round will start
         uint64 jurorsNumber;           // Jurors number for the next round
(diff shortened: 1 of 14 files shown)


contracts/disputes/DisputeManager.sol
```diff
@@ -242,14 +251,15 @@ contract DisputeManager is ControlledRecoverable, ICRVotingOwner, IDisputeManage
...
+        uint256 roundId = dispute.rounds.length - 1;
+        AdjudicationRound storage round = dispute.rounds[roundId];
         uint64 draftTermId = round.draftTermId;
         require(draftTermId <= currentTermId, ERROR_DRAFT_TERM_NOT_REACHED);
         bytes32 draftTermRandomness = clock.ensureCurrentTermRandomness();
 
         // Draft jurors for the given dispute and reimburse fees
         DraftConfig memory config = _getDraftConfig(draftTermId);
-        bool draftEnded = _draft(_disputeId, round, currentTermId, draftTermRandomness, config);
+        bool draftEnded = _draft(round, _buildDraftParams(_disputeId, roundId, currentTermId, draftTermRandomness, config));
```
bingen: Was this due to stack too deep issues?

contracts/registry/JurorsRegistry.sol
```diff
@@ -224,15 +226,21 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
     * @param _amount Amount of tokens to be added to the available balance of a juror
     */
     function assignTokens(address _juror, uint256 _amount) external onlyDisputeManager {
-        _updateAvailableBalanceOf(_juror, _amount, true);
+        if (_amount > 0) {
+            _updateAvailableBalanceOf(_juror, _amount, true);
+            emit JurorRewarded(_juror, _amount);
```
bingen: The `Reward` part of the event name looks like out of the scope of this function. We could use something more agnostic like `TokensAssigned`.

contracts/registry/JurorsRegistry.sol
```diff
@@ -224,15 +226,21 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
...
 
     /**
     * @notice Burn `@tokenAmount(self.token(), _amount)`
     * @param _amount Amount of tokens to be burned
     */
     function burnTokens(uint256 _amount) external onlyDisputeManager {
-        _updateAvailableBalanceOf(BURN_ACCOUNT, _amount, true);
+        if (_amount > 0) {
+            _updateAvailableBalanceOf(BURN_ACCOUNT, _amount, true);
+            emit BurnedTokens(_amount);
```
bingen: For consistency I would call this one `TokensBurned`.

contracts/registry/JurorsRegistry.sol
```diff
@@ -373,7 +385,7 @@ contract JurorsRegistry is ControlledRecoverable, IJurorsRegistry, ERC900, Appro
             tree.update(juror.id, nextTermId, _amount, false);
         }
 
-        emit JurorTokensCollected(_juror, _amount, nextTermId);
+        emit JurorSlashed(_juror, _amount, nextTermId);
```
bingen: I liked better the original name, as the function is not called `slash` but `collectTokens`.

contracts/disputes/DisputeManager.sol
facuspagnuolo: Yup :/

PR #251, opened 2020-02-10
Tests: Fix gas cost increases after Istambul hard-fork

facuspagnuolo:


test/lib/hex-sum-tree-gas.js
```diff
@@ -292,7 +292,7 @@ contract('HexSumTree', () => {
 
           context('with checkpoints', () => {
             const updateTimes = 100
-            const expectedCost = 895e3
+            const expectedCost = 1650e3
```
sohkai: Jesus, you can really tell this is SLOAD-heavy

PR #255, opened 2020-03-04
Update solidity-coverage to v0.7

### LINKED ISSUE #86: Add coverage measure tool

bingen:
Fixes #86

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,3 +43,9 @@ jobs:
     - stage: tests
       script: npm test test/treasury/court-treasury.js
       name: "Treasury"
+    - stage: tests
+      script npm run coverage
+      name: "Coverage"
+after_success:
+  - cat coverage/lcov.info | ./node_modules/.bin/coveralls
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,9 +43,8 @@ jobs:
     - stage: tests
       script: npm test test/treasury/court-treasury.js
       name: "Treasury"
-    - stage: tests
-      script npm run coverage
+    - stage: coverage
+      script: npm run coverage
       name: "Coverage"
 after_success:
   - cat coverage/lcov.info | ./node_modules/.bin/coveralls
-
--- a/test/lib/hex-sum-tree-gas.js
+++ b/test/lib/hex-sum-tree-gas.js
@@ -24,7 +24,7 @@ contract('HexSumTree', () => {
     })
   }
 
-  describe('gas costs', () => {
+  describe('gas costs [ @skip-on-coverage ]', () => {
     describe('insert', () => {
       context('non increasing height', () => {
         context('small tree', () => {
--- a/.solcover.js
+++ b/.solcover.js
@@ -4,9 +4,15 @@ const skipFiles = [
   'standards'
 ]
 
+const providerOptions = {
+  "total_accounts": 200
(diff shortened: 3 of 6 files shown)


facuspagnuolo:
I don't see a CI job for this, do you know why?

.travis.yml
```diff
@@ -43,3 +43,9 @@ jobs:
     - stage: tests
       script: npm test test/treasury/court-treasury.js
       name: "Treasury"
+    - stage: tests
+      script npm run coverage
```
sohkai:
```suggestion
    - stage: coverage
      script: npm run coverage
```

test/lib/hex-sum-tree-gas.js
```diff
@@ -24,7 +24,7 @@ contract('HexSumTree', () => {
     })
   }
 
-  describe('gas costs', () => {
+  describe('gas costs [ @skip-on-coverage ]', () => {
```
sohkai: This syntax is super super nice!

Do these fail though? solidity-coverage@0.7 shouldn't be increasing gas costs anymore, right?

.travis.yml
facuspagnuolo: this must be the reason why the CI didn't trigger a job for this PR... applying this suggestion to start the CI and see how it goes

test/lib/hex-sum-tree-gas.js
facuspagnuolo: interesting, let's give it a try

.travis.yml
bingen: Why wouldn’t it trigger the job? It’s the same stage as others. :thinking: 
It’s funny that I initially set this as `stage: coverage`, but then I saw that `lint` was also on `stage: tests` so I changed it to be consistent (I see coverage more like tests than linting).

test/lib/hex-sum-tree-gas.js
bingen: Yes, these were failing. Actually  I took that nice syntax from `solidity-coverage` documentation when I realized gas tests were failing.

.travis.yml
bingen: Oh, the `:`!!! :man_facepalming:

facuspagnuolo:
For some reason the `@skip-coverage` notation is not working

bingen:
> For some reason the @skip-coverage notation is not working

Do you mean in the CI? Because in locally it was working for me. I don't see gas tests in the coverage task either. Where are you seeing them?

facuspagnuolo (commented):
Let's see how we can handle the travis timeout, not sure if there is a way to instrument the code in batches

bingen:
> Let's see how we can handle the travis timeout, not sure if there is a way to instrument the code in batches

Being coverage I don't know how we can batch it. What I would do for now is to remove the travis job and merge this PR, so at least we can run coverage locally.

sohkai:
> Being coverage I don't know how we can batch it. What I would do for now is to remove the travis job and merge this PR, so at least we can run coverage locally.

We could do a quick experiment or two using lcov-merger (see aragon-apps) if we separate coverage runs to test separate files from each other.

facuspagnuolo:
Closing this for now

PR #264, opened 2020-05-01
Subscriptions: Implement trusted model

facuspagnuolo:
Based on a recent idea of having a more "trusted" or "trust-delegated" pattern for court subscription fees, I'm throwing this proposal of having an open subscription model where anyone is considered to be always subscribed but they can register payments with a reference jurors can use to determine if the court subscription fee was paid in order to rule or not a dispute.

This functionality will basically allow anyone to raise disputes to court paying the dispute fees, and jurors will then decide if they rule it or not. We could add some sort of advice to the dashboard to make clear if we found a payment reference for a dispute. To do that we will need to define some structure for the payment references ofc.


luisivan:
This feels like an interesting functionality although it also opens the door to more subjectivity and potentially less guarantees about being able to use the service.
So while I like the idea of abstracting the payment mechanism, we need to make sure it doesn't have unintended consequences. My main worry is that people wouldn't feel certainty about their cases being disputed, but I guess that it wouldn't be a problem since even if 2 out of 3 jurors don't think a payment mechanism is valid, you could eventually raise the case to the top and then it's up to the entire network to decide.

Thoughts @izqui @lkngtn?

bingen:
The idea sounds good, I only see one little problem: how to make sure all jurors are on the same page about the payment. Because if some jurors refuse to vote, but others don’t (i.e., they do vote), they could get slashed. So it’s like adding another layer of complexity to the adjudication process, the payment, which is actually independent of the dispute content.

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -452,36 +287,41 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
...
+            _token == ERC20(0)
+                ? require(msg.value == _amount, ERROR_TOKEN_DEPOSIT_FAILED)
+                : require(_token.safeTransferFrom(_from, address(this), _amount), ERROR_TOKEN_DEPOSIT_FAILED);
+        }
+    }
+
+    function _transfer(address payable _to, ERC20 _token, uint256 _amount) internal {
+        if (_amount > 0) {
+            _token == ERC20(0)
+                ? require(_to.send(_amount), ERROR_TOKEN_TRANSFER_FAILED) // solium-disable-line security/no-send
```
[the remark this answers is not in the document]
facuspagnuolo: Yup, the problem is that `transfer` does not allow you to specify an error message. I guess I based this on the implementation we are using for `Vault`, but we may want to discard the error message then

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -210,19 +137,10 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
...
-    /**
-    * @notice Pause sender subscriptions
-    */
-    function pause() external {
-        Subscriber storage subscriber = subscribers[msg.sender];
-        require(subscriber.subscribed, ERROR_SENDER_NOT_SUBSCRIBED);
-
-        subscriber.previousDelayedPeriods = uint64(_getDelayedPeriods(subscriber, _getCurrentPeriodId()));
-        subscriber.paused = true;
+        address payable juror = address(uint160(msg.sender));
```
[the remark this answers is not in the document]
facuspagnuolo: I was getting a compiler warning, that was the reason, we can review it again just in case

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -138,59 +91,34 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
...
-    function resume(uint256 _periods) external {
-        Subscriber storage subscriber = subscribers[msg.sender];
-        require(subscriber.paused, ERROR_SUBSCRIPTION_NOT_PAUSED);
-
-        _payFees(subscriber, msg.sender, msg.sender, _periods);
-
-        subscriber.paused = false;
-        subscriber.previousDelayedPeriods = 0;
+    function payFees(address _to, bytes calldata _reference) external payable {
+        _payFees(msg.sender, _to, _reference);
```
facuspagnuolo, line 103: Since we are going to add a new entry point here, let's rollback the reference field

contracts/subscriptions/ITransactionFeesOracle.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.5.8;
```
facuspagnuolo, line 1: I think we should remove these

test/helpers/utils/errors.js
```diff
@@ -68,25 +68,28 @@ const TREASURY_ERRORS = {
...
-  SUBSCRIPTION_NOT_PAUSED: 'CS_SUBSCRIPTION_NOT_PAUSED'
+  COURT_HAS_NOT_STARTED: 'CS_COURT_HAS_NOT_STARTED'
+}
+
+const TRANSACTIONS_FEES_ORACLE_ERRORS = {
+  ERROR_APP_NOT_SET: 'SFO_APP_NOT_SET',
+  ERROR_WRONG_TOKEN: 'SFO_WRONG_TOKEN',
+  ERROR_COURT_HAS_NOT_STARTED: 'SFO_COURT_HAS_NOT_STARTED',
+  ERROR_WRONG_TOKENS_LENGTH: 'SFO_WRONG_TOKENS_LENGTH',
+  ERROR_WRONG_AMOUNTS_LENGTH: 'SFO_WRONG_AMOUNTS_LENGTH'
```
facuspagnuolo, line 92: Let's not forget to update these

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -550,40 +505,16 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
     * @param _feeToken New ERC20 token to be used for the subscription fees
     */
     function _setFeeToken(ERC20 _feeToken) internal {
-        require(isContract(address(_feeToken)), ERROR_FEE_TOKEN_NOT_CONTRACT);
+        require(address(_feeToken) == ETH || isContract(address(_feeToken)), ERROR_FEE_TOKEN_NOT_CONTRACT);
 
         if (accumulatedGovernorFees > 0) {
+            // There’s no re-entrancy issue here: governor should be trusted
             _transferFeesToGovernor();
```
facuspagnuolo, line 512: ⚠️ This one was carried from before but we should fix it. Note that `_transferFeesToGovernor` relies on `currentFeeToken` while the governor fees totalizator is using the current period token cause it is being added when users pay their fees. Similar to what we are doing with the `DisputeManager` we should always use the same config for the entire period. Let's move the governor fees totalizator to the period information and fix the transfer method to work with period IDs similar to how the jurors do

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -451,37 +369,58 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
...
         accumulatedGovernorFees = accumulatedGovernorFees.add(governorFee);
 
         // Update collected fees for the jurors
-        uint256 collectedFees = amountToPay.sub(governorFee);
-        period.collectedFees = period.collectedFees.add(collectedFees);
-
-        // Periods are measured in Court terms. Since Court terms are represented in uint64, we are safe to use uint64 for period ids too.
-        _subscriber.lastPaymentPeriodId = uint64(newLastPeriodId);
+        uint256 collectedFees = _feeAmount.sub(governorFee);
+        _period.collectedFees = _period.collectedFees.add(collectedFees);
```
facuspagnuolo, line 395: ⚠️ We are mixing tokens here. Currently, we have multiple fee tokens, one for the jurors, and one per app. The only way I find we can avoid this is to make a few changes in `_setAppFees`:
1. Schedule the change for the next period as we are doing with the subscriptions fee token
2. Add a check to make sure the given fee token is the same as the next period fee token.

Note that (2) will ensure (1) by default and basically force the governor to use `setFeeToken` to change the app fee token.

test/subscriptions/aragon-app-fees-cashier.js
```diff
@@ -161,16 +169,88 @@ contract('Aragon App Fees Cashier', ([_, governor, subscriber, fakeToken]) => {
...
+          receipt = await aragonAppFeesCashier.payAppFees(appId, EMPTY_DATA, { from: subscriber })
+        }
+
+        const finalBalance = await getBalance(isEth)
+
+        assertBn(finalBalance, initialBalance.add(amount), 'amount doesn’t match')
+
+        assertAmountOfEvents(receipt, SUBSCRIPTIONS_EVENTS.APP_FEE_PAID)
+        assertEvent(receipt, SUBSCRIPTIONS_EVENTS.APP_FEE_PAID, { by: subscriber, appId, data: EMPTY_DATA })
+      })
```
facuspagnuolo, line 225: Based on what I explained in the contract, let's add some more specific scenarios here, similar to what we have in `subscriptions-payments`, `subscriptions-jurors`, and `subscriptions-donations`, to make sure the math adds correctly for the jurors and governor shares.

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -73,62 +70,32 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
...
-        uint256 governorFee
-    );
-    event FeesDonated(address indexed payer, uint256 indexed periodId, uint256 amount);
-    event FeesClaimed(address indexed juror, uint256 indexed periodId, uint256 jurorShare);
-    event GovernorFeesTransferred(uint256 amount);
-    event FeeTokenChanged(address previousFeeToken, address currentFeeToken);
+    // List of fees per appId
+    mapping (bytes32 => AppFee) internal appFees;
+
+    event FeesPaid(address indexed subscriber, uint256 indexed periodId, ERC20 feeToken, uint256 feeAmount, bytes ref);
```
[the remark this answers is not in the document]
bingen: Yes, but those events are independent: this one will happen when calling `payFees` while the other one will happen when calling `payAppFees`.

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -451,37 +369,58 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
...
+        }
+    }
+
+    function _transfer(address payable _to, ERC20 _token, uint256 _amount) internal {
+        if (_amount == 0) {
+            return;
+        }
+
+        if (address(_token) == ETH) {
+            (bool success, ) = _to.call.value(_amount)(""); // solium-disable-line security/no-call-value
```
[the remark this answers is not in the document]
bingen: Without it it doesn’t compile:
```
Wrong argument count for function call: 0 arguments given but expected 1. This function requires a single bytes argument. Use "" as argument to provide empty calldata.
```

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -597,17 +528,33 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
...
+    * @notice Unset fees for app with id `_appId`
+    * @param _appId Id of the app
     */
-    function _setResumePrePaidPeriods(uint256 _resumePrePaidPeriods) internal {
-        // Check resume resume pre-paid periods it not above the number of allowed pre payment periods. Since these periods are always paid in
-        // advance, we must make sure there won't be users covering too many periods in the future to avoid skipping fee changes or
-        // excluding many jurors from their corresponding rewards.
-        require(_resumePrePaidPeriods <= prePaymentPeriods, ERROR_RESUME_PRE_PAID_PERIODS_TOO_BIG);
+    function _unsetAppFee(bytes32 _appId) internal {
+        require(appFees[_appId].set, ERROR_APP_FEE_NOT_SET);
```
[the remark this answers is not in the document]
bingen: I only see it here and in another internal function, called like the one proposed here `getAppFee`. Here we cannot use this because then `delete` doesn’t work:
```
Unary operator delete cannot be applied to type struct CourtSubscriptions.AppFee storage pointer
```

contracts/subscriptions/CourtSubscriptions.sol
bingen: Good point!! But it makes me think that having different tokens per app is unnecessary complexity. I’m thinking that it would make UX way worse for both users paying fees (and having to stake in a potential lot of staking pools) and for users claiming fees. So, what do you think about using in this implementation `currentFeeToken` for all apps? We could still keep the `IAragonAppFeesCashier` as it is, in case we realize later than for some use case it’s needed to have a different token for a different app.

contracts/subscriptions/CourtSubscriptions.sol
bingen: Good catch!!! I’m going to open a new issue for this to not mix things up.

bingen: https://github.com/aragon/aragon-court/issues/273

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -223,27 +222,34 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
     // IAragonAppFeesCashier interface
 
     /**
-    * @notice Set fees for app with id `_appId` to @tokenAmount(`_token`, `_amount`)
```
facuspagnuolo: We can probably leave this one, the transaction will revert if a wrong token is being used

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -272,21 +278,23 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
     * @param _data Extra data for context of the payment
     */
     function payAppFees(bytes32 _appId, bytes calldata _data) external payable {
-        (ERC20 feeToken, uint256 feeAmount) = _getAppFee(_appId);
+        uint256 feeAmount = _getAppFee(_appId);
```
facuspagnuolo, line 281: Note that we are not using a checkpointed value here... this could be weird for the governor if they change the token, they will need to update the app fees amount right when the period starts. Although, I don't see a way to checkpoint these fees on each period

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -375,7 +383,11 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
     * @return Amount of fee tokens
     */
     function getAppFee(bytes32 _appId) external view returns (ERC20 token, uint256 amount) {
-        return _getAppFee(_appId);
+        uint256 currentPeriodId = _getCurrentPeriodId();
+        Period storage period = periods[currentPeriodId];
```
facuspagnuolo, line 387: Let's use `_getCurrentPeriod` here

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -223,27 +222,34 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
...
+
+        _setAppFee(_appId, feeToken, _amount);
     }
 
     /**
     * @notice Set fees for apps with ids `_appIds`
     * @param _appIds Id of the apps
-    * @param _tokens Token for the fees for each app
+    * @param _tokens Token for the fees for each app (must be an empty array, as we are using the global token)
     * @param _amounts Amount of fee tokens for each app
```
facuspagnuolo, line 242: Same here about the note regarding the fee amount change

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -223,27 +222,34 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
     // IAragonAppFeesCashier interface
 
     /**
-    * @notice Set fees for app with id `_appId` to @tokenAmount(`_token`, `_amount`)
+    * @notice Set fees for app with id `_appId` to `_amount`
     * @param _appId Id of the app
-    * @param _token Token for the fee
+    * @param _token Token for the fee, must be the same as the current period one
     * @param _amount Amount of fee tokens
```
facuspagnuolo, line 228: We can probably leave a note here explaining that a fee amount change is executed immediately, there is no way to schedule these for the next period.

contracts/subscriptions/CourtSubscriptions.sol
bingen: Good point!! Checkpointing `appFees` sounds like too much overhead, but changing the token would definitely be a problem for appFees. Maybe we could keep a list of enabled apps and disable them (`appFees[appId].set = false`) when changing the token, to force the governor to update them.

PR #268, opened 2020-05-28
Arbitrable: Add arbitrator to evidence event

facuspagnuolo:


contracts/arbitration/IArbitrable.sol
```diff
@@ -18,12 +18,13 @@ contract IArbitrable is ERC165 {
...
     /**
     * @dev Emitted when new evidence is submitted for the IArbitrable instance's dispute
+    * @param arbitrator IArbitrator submitting the evidence for
     * @param disputeId Identification number of the dispute receiving new evidence
     * @param submitter Address of the account submitting the evidence
     * @param evidence Data submitted for the evidence of the dispute
     * @param finished Whether or not the submitter has finished submitting evidence
     */
-    event EvidenceSubmitted(uint256 indexed disputeId, address indexed submitter, bytes evidence, bool finished);
+    event EvidenceSubmitted(IArbitrator indexed arbitrator, uint256 indexed disputeId, address indexed submitter, bytes evidence, bool finished);
```
sohkai, line 27: I'm wondering about backwards compatibility with the mainnet Court instance and the subgraph.

But in reality, this shouldn't be a problem, right? In that subgraph, we're not actually listening to any Aribtrables?

facuspagnuolo: We are but only to the ones that submit disputes to the court. We will need to update the subgraph to make sure the arbitrator is correct, otherwise we could be rendering evidence for another court for example. I'll take care of that once we release this change :)

sohkai: Have we considered using an `ArbitratorChanged` event instead?

It would save a bit of gas and be less of a breaking change, event wise, but assumes an Arbitratable is only hooked up to a single Arbitrator at a time (perhaps not something we want to commit to).

But regardless, I still like what we're currently doing better.

PR #269, opened 2020-06-02
Docs: Update testing guide with new Rinkeby instances

facuspagnuolo:


docs/8-testing-guide/readme.md
```diff
@@ -33,13 +33,13 @@ This testing instance mirrors the instance deployed to Mainnet, same terms durat
 
 - Network: Rinkeby
```
sohkai, line 34: On this file, I notice that we have usability; should we switch to staging instead?

PR #270, opened 2020-06-18
New subscription fees

bingen:


contracts/subscriptions/ISubscriptionFeesOracle.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.5.8;
+
+import "../lib/os/ERC20.sol";
+
+
+interface ISubscriptionFeesOracle {
```
facuspagnuolo, line 6: I know this comes from the spec I proposed, but now that think of it, it probably makes more sense to call it `ITransactionFeesOracle`

contracts/subscriptions/SubscriptionFeesOracle.sol
```diff
@@ -0,0 +1,48 @@
...
+
+        AppFee storage appFee = appFees[_appId];
+        appFee.token = _token;
+        appFee.amount = _amount;
+    }
+
+    function getFee(bytes32 _appId) external view returns (ERC20 token, uint256 amount, address beneficiary) {
+        require(_appId != bytes32(0), ERROR_APP_ID_ZERO);
+        // TODO: should we check this?
+        require(_getCurrentTermId() > 0, ERROR_COURT_HAS_NOT_STARTED);
```
facuspagnuolo, line 42: Maybe it seems an overhead for a query method, the payment will fail if the court hasn't started tho, I'd remove it to be honest

contracts/subscriptions/SubscriptionFeesOracle.sol
```diff
@@ -0,0 +1,48 @@
...
+        require(_appId != bytes32(0), ERROR_APP_ID_ZERO);
+        require(address(_token) == ETH || isContract(address(_token)), ERROR_WRONG_TOKEN);
+
+        AppFee storage appFee = appFees[_appId];
+        appFee.token = _token;
+        appFee.amount = _amount;
+    }
+
+    function getFee(bytes32 _appId) external view returns (ERC20 token, uint256 amount, address beneficiary) {
+        require(_appId != bytes32(0), ERROR_APP_ID_ZERO);
```
facuspagnuolo, line 40: Same here, I think it is not necessary to fail for this case, we can simply return 0 values

contracts/subscriptions/SubscriptionFeesOracle.sol
```diff
@@ -0,0 +1,48 @@
...
+    string private constant ERROR_APP_ID_ZERO = "SFO_APP_ID_ZERO";
+    string private constant ERROR_WRONG_TOKEN = "SFO_WRONG_TOKEN";
+    string private constant ERROR_COURT_HAS_NOT_STARTED = "SFO_COURT_HAS_NOT_STARTED";
+
+    struct AppFee {
+        ERC20 token;
+        uint256 amount;
+    }
+
+    mapping (bytes32 => AppFee) appFees;
```
facuspagnuolo, line 20: let's declare this internal

contracts/subscriptions/SubscriptionFeesOracle.sol
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+    * @dev Initialize court subscription fees oracle
+    * @param _controller Address of the controller
+    */
+    constructor(Controller _controller) Controlled(_controller) public {
+        // solium-disable-previous-line no-empty-blocks
+    }
+
+    function setFee(bytes32 _appId, ERC20 _token, uint256 _amount) external onlyConfigGovernor {
```
facuspagnuolo, line 30: what do you think about having another setter for a bunch of apps?

contracts/subscriptions/SubscriptionFeesOracle.sol
```diff
@@ -0,0 +1,48 @@
...
+        require(_appId != bytes32(0), ERROR_APP_ID_ZERO);
+        require(address(_token) == ETH || isContract(address(_token)), ERROR_WRONG_TOKEN);
+
+        AppFee storage appFee = appFees[_appId];
+        appFee.token = _token;
+        appFee.amount = _amount;
```
facuspagnuolo, line 36: Let's emit an event here so we can have a way to retrieve historic information

contracts/subscriptions/SubscriptionFeesOracle.sol
bingen: In my opinion, returning 0 values is not a good idea generally speaking, as it can lead to bugs like [this one](https://github.com/aragonone/court-presale/pull/6). 
Besides, it’s more consistent with the “out of bounds” pattern we usually follow (see [here](https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L240) for instance)
Ideally I would like to fail for any non-valid case, not only this one, so I’m going to make some changes in that direction.

contracts/subscriptions/TransactionFeesOracle.sol
```diff
@@ -0,0 +1,113 @@
...
+    struct AppFee {
+        bool set;
+        ERC20 token;
+        uint256 amount;
+    }
+
+    mapping (bytes32 => AppFee) internal appFees;
+
+    event AppFeeSet(bytes32 indexed _appId, ERC20 _token, uint256 _amount);
+    event AppFeeUnset(bytes32 indexed _appId);
```
facuspagnuolo, line 26: Nitpick: I think we usually avoid the `_` prefix for the events

contracts/subscriptions/TransactionFeesOracle.sol
```diff
@@ -0,0 +1,113 @@
...
+    /**
+    * @dev Initialize court subscription fees oracle
+    * @param _controller Address of the controller
+    */
+    constructor(Controller _controller) Controlled(_controller) public {
+        // solium-disable-previous-line no-empty-blocks
+    }
+
+    /**
+    * @notice Set fees for app with id `_appId` to `_amount` of `_token` tokens
```
facuspagnuolo, line 37: Let's use the `@tokenAmount` radspec notation

contracts/subscriptions/TransactionFeesOracle.sol
```diff
@@ -0,0 +1,113 @@
...
+    * @notice Get fees for app with id `_appId`
+    * @param _appId Id of the app
+    * @return Token for the fees
+    * @return Amount of fee tokens
+    * @return Beneficiary to send the fees to
+    */
+    function getFee(bytes32 _appId) external view returns (ERC20 token, uint256 amount, address beneficiary) {
+        AppFee storage appFee = appFees[_appId];
+
+        require(appFee.set, ERROR_APP_ID_NOT_SET);
```
facuspagnuolo, line 89: I think I would simply return zero in this case, any reason to fail instead?

contracts/arbitration/IArbitrator.sol
```diff
@@ -40,4 +40,11 @@ interface IArbitrator {
     * @return feeAmount Total amount of fees that must be allowed to the recipient
     */
     function getSubscriptionFees(address _subscriber) external view returns (address recipient, ERC20 feeToken, uint256 feeAmount);
+
+    /**
+    * @dev Tell address of a module based on a given ID
+    * @param _id ID of the module being queried
+    * @return Address of the requested module
+    */
+    function getModule(bytes32 _id) external view returns (address);
```
facuspagnuolo, line 49: Instead of adding this to the `IArbitrator` I will denote that the Agreement app relies on the AragonCourt interface instead. WDYT?

contracts/subscriptions/ITransactionFeesOracle.sol
```diff
@@ -0,0 +1,12 @@
...
+
+import "../lib/os/ERC20.sol";
+
+
+interface ITransactionFeesOracle {
+    function setFee(bytes32 appId, ERC20 token, uint256 amount) external;
+    function setFees(bytes32[] calldata _appIds, ERC20[] calldata _tokens, uint256[] calldata _amounts) external;
+    function unsetFee(bytes32 _appId) external;
+    function unsetFees(bytes32[] calldata _appIds) external;
+    function getFee(bytes32 appId) external view returns (ERC20, uint256, address);
```
facuspagnuolo, line 11: Let's use the `_` prefix for all the args here

contracts/arbitration/IArbitrator.sol
bingen: I don’t think I follow, [here](https://github.com/aragon/aragon-apps/pull/1176/files#diff-9ebddb6de87cd26830aa77a8db008cedR712) we have `IArbitrator`.

bingen: Discussion continued [here](https://github.com/aragon/aragonOS/pull/586#discussion_r445620075)

bingen:
Replaced by #272

PR #275, opened 2020-07-23
Remove ETH from AragonAppFeesCashier

bingen:


bingen:
> We should revert the changes on the subscription module we did to support ETH as well.

Yes, yes, I started from an outdated `development` commit (hence the conflicts). Pushing again in a few.

PR #277, opened 2020-08-11
Test: Adopt new contract test helpers

facuspagnuolo:


test/registry/jurors-registry-slashing.js
```diff
@@ -312,26 +308,26 @@ contract('JurorsRegistry', ([_, juror, secondJuror, thirdJuror, anyone]) => {
...
-            assertEvent({ logs }, REGISTRY_EVENTS.JUROR_TOKENS_COLLECTED, { juror, amount, effectiveTermId: termId.add(bn(1)) })
+            assertEvent({ logs }, REGISTRY_EVENTS.JUROR_TOKENS_COLLECTED, { expectedArgs: { juror, amount, effectiveTermId: termId.add(bn(1)) } })
           })
         }
 
         it('does not process deactivation requests', async () => {
           const receipt = await disputeManager.collect(juror, amount)
 
-          assertAmountOfEvents(receipt, REGISTRY_EVENTS.JUROR_DEACTIVATION_PROCESSED, 0)
+          assertAmountOfEvents(receipt, REGISTRY_EVENTS.JUROR_DEACTIVATION_PROCESSED, { expectedAmount: 0 })
```
sohkai, line 330: Oh interesting, it might be worth having a small wrapper around this to assert no events!

test/registry/jurors-registry-slashing.js
```diff
@@ -115,7 +111,7 @@ contract('JurorsRegistry', ([_, juror, secondJuror, thirdJuror, anyone]) => {
             await registry.mockNextDraft(draftedJurors, draftedWeights)
 
             // Draft and make sure mock worked as expected
-            const receipt = await disputeManager.draft(EMPTY_RANDOMNESS, 1, 0, 10, 10, DRAFT_LOCK_PCT)
+            const receipt = await disputeManager.draft(ZERO_BYTES32, 1, 0, 10, 10, DRAFT_LOCK_PCT)
```
sohkai, line 114: I think in this case, keeping the "empty randomness" could be useful—it signals what the `draft()` argument is for :)

test/court/controller/controller-clock.js
```diff
@@ -113,7 +109,7 @@ contract('Controller', ([_, someone, configGovernor]) => {
 
           assertBn(startTime, firstTermStartTime.add(termDuration.mul(bn(transition - 1))), `start time for term ${termId} does not match`)
           assertBn(randomnessBN, currentBlockNumber.add(bn(1)), `randomness block number for term ${termId} should be the next block number`)
-          assert.equal(randomness, EMPTY_RANDOMNESS, `randomness for term ${termId} should not be computed`)
+          assert.equal(randomness, ZERO_BYTES32, `randomness for term ${termId} should not be computed`)
```
sohkai, line 112: And similarly here :)

test/helpers/wrappers/court.js
```diff
@@ -59,416 +59,412 @@ const DEFAULTS = {
...
-
-  class CourtHelper {
-    constructor(web3, artifacts) {
-      this.web3 = web3
-      this.artifacts = artifacts
-    }
+class CourtHelper {
+  constructor() {
+    this.web3 = getWeb3()
+    this.artifacts = getArtifacts()
```
sohkai, line 65: Nice! I never thought using the global configuration mechanism from the contract helpers like this, but I like it!

Technically you could still accept the custom `web3` and `artifacts` if you needed to, but I don't see a particular reason why we would want to keep them since this helper is internal to our tests here :).

PR #278, opened 2020-08-11
Chore: Migrate to yarn + GH actions

facuspagnuolo:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,157 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
sohkai, line 7: Nitpick: for yaml we are generally avoiding an indent for the `-`. So this would look like:

```
steps:
- uses: ...
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Install node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 10
+      - name: Install
+        run: yarn --frozen-lockfile
+      - name: Treasury
+        run: yarn test test/treasury/court-treasury.js
+    env:
```
sohkai, line 156: Nitpick: you can [move `env` up to the root of the definition file](https://github.com/aragon/aragon-apps/pull/1223/files#diff-d3711c93388c5b7cc98a29a581481d10R36), so you don't need to repeat it for each job :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,157 @@
...
+on: [push, pull_request]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 10
```
sohkai, line 11: Are we still forced to use node 10? If so, hopefully it's just solidity-coverage :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,157 @@
+name: CI
+on: [push, pull_request]
```
sohkai, line 2: To avoid duplicate runs, we can set [`push` to be just on `master`](https://github.com/aragon/aragon-apps/pull/1223/files#diff-d3711c93388c5b7cc98a29a581481d10R5)

.github/workflows/ci.yml
facuspagnuolo: I'll give it a try in a separate PR :)

PR #280, opened 2020-08-12
Chore: Add solidity coverage v7

facuspagnuolo:


package.json
```diff
@@ -10,14 +10,13 @@
...
   "scripts": {
-    "compile": "truffle compile",
     "lint": "yarn lint:solidity && yarn lint:js",
     "lint:js": "eslint ./test ./scripts ./migrations",
     "lint:solidity": "solium --dir ./contracts",
     "test": "yarn test:ganache-cli",
     "test:gas": "GAS_REPORTER=true yarn test",
     "test:ganache-cli": "./node_modules/@aragon/contract-helpers-test/scripts/ganache-cli.sh",
-    "coverage": "SOLIDITY_COVERAGE=true yarn test:ganache-cli",
+    "coverage": "npx truffle run coverage",
```
sohkai, line 19: I think we need to add a plugin to the truffle config now :).

That being said... how long do you think this is going to take haha?

facuspagnuolo: A lot actually, I think probably it does worth fixing it but not sure if it helps adding it to the CI

PR #282, opened 2020-09-09
Subscriptions: Internal audit

facuspagnuolo:
Addressing @sohkai's findings. I recommend reviewing commit by commit


test/subscriptions/aragon-app-fees-cashier.js
```diff
@@ -165,7 +164,7 @@ contract('Aragon App Fees Cashier', ([_, governor, subscriber]) => {
     context('when the court hasn’t started', () => {
       it('reverts', async () => {
         await assertRevert(aragonAppFeesCashier.setAppFee(appId, feeToken.address, amount, { from: governor }), SUBSCRIPTIONS_ERRORS.COURT_HAS_NOT_STARTED)
-        await assertRevert(aragonAppFeesCashier.payAppFees(appId, EMPTY_BYTES, { from: subscriber }), SUBSCRIPTIONS_ERRORS.APP_FEE_NOT_SET)
+        await assertRevert(aragonAppFeesCashier.payAppFees(appId, EMPTY_BYTES, { from: subscriber }), SUBSCRIPTIONS_ERRORS.COURT_HAS_NOT_STARTED)
```
sohkai, line 167: This is also a much better error than before :)

test/subscriptions/aragon-app-fees-cashier.js
```diff
@@ -174,9 +173,14 @@ contract('Aragon App Fees Cashier', ([_, governor, subscriber]) => {
...
-        it('reverts', async () => {
-          await assertRevert(aragonAppFeesCashier.payAppFees(appId, EMPTY_BYTES, { from: subscriber }), SUBSCRIPTIONS_ERRORS.APP_FEE_NOT_SET)
+      context('when the app fee has not been set', () => {
+        it('ignores the payment', async () => {
+          const initialBalance = await feeToken.balanceOf(aragonAppFeesCashier.address)
+
+          await aragonAppFeesCashier.payAppFees(appId, EMPTY_BYTES, { from: subscriber })
+
+          const finalBalance = await feeToken.balanceOf(aragonAppFeesCashier.address)
+          assertBn(finalBalance, initialBalance, 'subscription balance does not match')
```
sohkai, line 183: Should we also assert that the `AppFeePaid` event isn't logged?

Although I do wonder if we shouldn't log the event anyway... or should we just assume that if no event is associated, the app fees haven't been paid? In the case of Agreements, we also send extra data to be logged as well.

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -165,6 +165,12 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
     * @param _feeToken New ERC20 token to be used for the subscription fees
     */
     function setFeeToken(ERC20 _feeToken) external onlyConfigGovernor {
+        // Ensure the current period has its fee token set if the court has started
```
sohkai:
```suggestion
        // If the court has started, ensure the fee token is set for the current period before changing it
```

sohkai (approved):
Looking good!

test/subscriptions/aragon-app-fees-cashier.js
facuspagnuolo: Let's log it as well

contracts/subscriptions/CourtSubscriptions.sol
```diff
@@ -232,19 +235,21 @@ contract CourtSubscriptions is ControlledRecoverable, TimeHelpers, ISubscription
 
     /**
     * @notice Pay fees corresponding to a new action in app with id `appId`
-    * @dev To be called by Agreements. It needs a pre-approval of tokens
+    * @dev To be called by Agreements. If a fee is defined, it requires the sender to have pre-approved tokens.
```
sohkai:
```suggestion
    * @dev To be called by the eventual dispute creator (usually an IArbitrable). If a fee is defined, it requires the sender to have pre-approved tokens.
           We expect the caller to submit the correct appId here, and leave it to jurors to verify that the correct fees were paid ahead of disputes.
```

PR #283, opened 2020-09-18
Chore: Migrate to buidler

facuspagnuolo:


.gitignore
```diff
@@ -2,11 +2,12 @@
 
 # Dependencies
 node_modules/
-package-lock.json
```
sohkai, line 5: Should we leave this in the gitignore, since people shouldn't be using npm?

PR #286, opened 2020-10-12
Chore: Organize repo content

facuspagnuolo:
Implementing new structure:
```
.
|- docs
|- packages
    |- evm                  // smart contract code
    |- subgraph             // subgraph code
    |- deployment           // deployment tooling (TODO)
```


packages/evm/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "eslint-config-standard": "^12.0.0",
+    "eslint-plugin-import": "^2.8.0",
+    "eslint-plugin-node": "^7.0.1",
+    "eslint-plugin-promise": "^4.0.1",
+    "eslint-plugin-standard": "^4.0.0",
+    "eslint-plugin-unused-imports": "^0.1.0",
+    "solidity-coverage": "^0.7.9",
+    "solium": "^1.2.5",
+    "web3": "^1.2.11",
+    "web3-eth-abi": "1.0.0-beta.33"
```
sohkai, line 54: We can probably bump this up to be the same version as web3@1.2.11

packages/evm/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@aragon/protocol-evm",
+  "version": "1.5.0",
```
sohkai, line 3: Haha no opposition but maybe we should go with "protocol v1" (even if it is technically a v1.5).

.github/workflows/subgraph_deploy_testnet.yml
```diff
@@ -0,0 +1,25 @@
+name: Subgraph testnets deployment
+
+on:
+  push:
+    branches:
+    - 'master'
```
sohkai, line 6: We can also do this by path, so that only changes to the subgraph package will trigger this

.github/workflows/subgraph_deploy_mainnet.yml
```diff
@@ -0,0 +1,21 @@
+name: Subgraph mainnet deployment
+
+on:
+  push:
+    tags:
+    - v*
```
sohkai, line 6: Thinking aloud: would this force us to make subgraph changes as part of any protocol change?

I think this is a good idea!

.github/workflows/ci_ethereum.yml
```diff
@@ -138,6 +148,7 @@ jobs:
     - name: Install
       run: yarn --frozen-lockfile
     - name: Treasury
-      run: yarn test test/treasury/court-treasury.js
+      run: yarn workspace @aragon/protocol-evm test:treasury
```
sohkai, line 151: Oh interesting, does the `workspace` part allow us to select a specific inner package and run the script in its context?

.github/workflows/ci_ethereum.yml
```diff
@@ -18,8 +18,9 @@ jobs:
     - name: Install
```
sohkai, line 18: I'd also use the `path` specifier for this workflow file :)

sohkai (commented):
Couple quick notes but looking good!

.github/workflows/subgraph_deploy_testnet.yml
facuspagnuolo: If we do this by path aren't we going to have different branches trying to be deployed even if they are still WIP?

.github/workflows/ci_ethereum.yml
facuspagnuolo: yea!

.github/workflows/subgraph_deploy_mainnet.yml
facuspagnuolo: I think so :)

facuspagnuolo:
Closing in favor of https://github.com/aragon/protocol/pull/1

.github/workflows/subgraph_deploy_testnet.yml
sohkai: Ah no, I meant the combination of master branch + path selector. That way, only merges to master with changes to the subgraph package would cause a subgraph deployment, rather than every merge to master.

